PR #9, opened 2025-07-25
Support remote browser window

zizhengtai:


src/narada/window.py
```diff
@@ -187,5 +170,44 @@ async def dispatch_request(
...
 
+class LocalBrowserWindow(BaseBrowserWindow):
+    _config: BrowserConfig
+    _context: BrowserContext
+
+    def __init__(
+        self,
+        *,
+        api_key: str,
+        id: str,
```
amirgholami, line 181: ditto here

PR #15, opened 2025-08-06
Add extension actions API

zizhengtai:
This enables a modular approach to building more complex custom workflows/agents. Example usage added to `examples/complex_workflow.py`.

### DIFF
--- a/src/narada/__init__.py
+++ b/src/narada/__init__.py
@@ -16,9 +16,6 @@ from narada.window import (
     ResponseContent,
 )
 
-__version__ = "0.1.11"
-
-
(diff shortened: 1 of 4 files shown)


src/narada/__init__.py
```diff
@@ -16,9 +16,6 @@
     ResponseContent,
 )
 
-__version__ = "0.1.11"
```
zizhengtai, line 19: There is actually no reason we need this. Modern Python can use `importlib.metadata.version("narada")` to get this information.

amirgholami (approved):

PR #32, opened 2025-09-05
Navigate to the `customToken` URL when spinning up a Narada instance

abrahmasandra:
Add the `customToken` that is retrieved from the `apiKey` to the URL when initiating the Narada instance. This allows the frontend to sign in automatically rather than waiting for user input to sign in.

### DIFF
--- a/packages/narada/src/narada/client.py
+++ b/packages/narada/src/narada/client.py
@@ -8,6 +8,7 @@ import sys
 from dataclasses import dataclass
 from typing import Any
 from uuid import uuid4
+from http import HTTPStatus
 
 from narada_core.errors import (
     NaradaExtensionMissingError,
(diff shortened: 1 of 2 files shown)


packages/narada/src/narada/client.py
```diff
@@ -84,6 +92,43 @@ async def __aexit__(self, *args: Any) -> None:
...
+
+        The `base_url` may be provided (for example `config.backend_api_base_url`);
+        otherwise the `NARADA_API_BASE_URL` env var is used with a sensible default.
+        Raises `NaradaTimeoutError` on timeouts and maps 504 to `NaradaTimeoutError`.
+        HTTP errors are raised as `httpx` exceptions.
+        """
+        url = f"{base_url.rstrip('/')}/auth/custom-token"
+
+        try:
+            async with httpx.AsyncClient() as client:
```
zizhengtai, line 109: we use `aiohttp` in this codebase

PR #85, opened 2026-04-09
[SDK] Attachment variables

abrahmasandra:


packages/narada/src/narada/window.py
```diff
@@ -132,6 +160,54 @@ async def upload_file(self, *, file: IO) -> File:
 
         return File(key=object_key)
 
+    async def _normalize_input_variables(
+        self, *, input_variables: dict[str, Any]
```
zizhengtai:
```suggestion
        self, *, input_variables: Mapping[str, Any]
```
Use generic types for input values and concrete types for output

packages/narada/src/narada/window.py
```diff
@@ -65,6 +66,28 @@
 _ResponseModel = TypeVar("_ResponseModel", bound=BaseModel)
 
 
+class _InputVariableFileReference(BaseModel):
+    key: str
+    name: str
+
+
+_JsonPrimitive = str | int | float | bool | None
```
zizhengtai:
```suggestion
type _JsonPrimitive = str | int | float | bool | None
```
Ditto for other type aliases below

zizhengtai (approved):

zizhengtai (approved):

PR #87, opened 2026-04-17
Add Python agent run trace

xTRam1:
## Summary

Adds structured tracing for custom Python agents. The SDK emits events for every stdout/stderr line, sub-agent call, extension action, and side-effect; the frontend ([NaradaAI/frontend#1592](https://github.com/NaradaAI/frontend/pull/1592)) consumes these events and renders them on the observability dashboard.

## What changes

**`narada-core`** — new Pydantic models
- `PythonAgentRunTrace` (step_type: `pythonAgentRun`) wrapping a run's full execution
- `PythonTraceEvent` discriminated union: `stdout`, `stderr`, `subAgentCall`, `extensionAction`, `sideEffect`
- Added to `ApaStepTrace`; `parse_action_trace` round-trips the new shape

**`narada-pyodide`** — chokepoint-based instrumentation so adding a new SDK method needs no trace-layer changes
- `window.py:dispatch_request()` wrapped → every `window.agent(...)` emits a `subAgentCall` event (covers success / error / timeout)
- `window.py:_run_extension_action()` wrapped → every extension action emits an `extensionAction` event
- `utils.download_file` / `utils.render_html` → `sideEffect` events
- All events flow through `_narada_emit_trace_event(json_str)`, a Pyodide global the frontend registers

**Payload safety** — per-action summaries in `_trace.summarize_request` bound the persisted JSON size (hard caps on free-form strings; structured subsets for `write_google_sheet`, screenshot, `agentic_selector`, etc.).

## Version bumps (coupled)

- `narada-core`: 0.0.17 → **0.0.18**
- `narada-pyodide`: 0.0.43 → **0.0.45a2**
- `narada`: 0.1.42 → **0.1.43** (repin `narada-core==0.0.18` only, no code change)

Why bump `narada` too: if `narada-core@0.0.18` (new variant) shipped without a matching `narada` bump, external Python SDK users would silently break on `parse_action_trace` ValidationError the moment a response they polled contained a nested Python custom agent run.

## Example

This custom Python agent:

```python
from narada import Agent, LocalBrowserWindow
import narada

window = LocalBrowserWindow()
await window.go_to_url(url="https://example.com", new_tab=True)

narada.render_html("<h1>Trace ordering test</h1>")
print("[A] after render_html")

narada.download_file("ordering-test.txt", "x")
print("[B] after download_file")

raise ValueError("intentional final traceback")
```

emits this trace (rendered by the frontend PR):

```json
{
  "step_type": "pythonAgentRun",
  "status": "error",
[...]

PR #93, opened 2026-04-24
feat(sdk): add reasoning effort to the Core Agent

xTRam1:
## Summary

Lets SDK callers opt into GPT-5.2's reasoning levels on the Core Agent — `none / low / medium / high` — mirroring the picker the web UI added on v2 of the agent step. The wire field is the same `reasoningMode` on `NaradaGenerationRequest` that the backend (`caddie/cloud/src/ecs/ai/agents/core_agent.py:46-66`) already honors only for `CoreAgent`; this PR just exposes it on the SDK surface.

> ⚠️ **Stacked on top of [#87](https://github.com/NaradaAI/narada-python-sdk/pull/87).** Base branch is `lerdogan/python-agent-trace`, not `main`. Once #87 merges, GitHub auto-rebases this onto `main`.

### New enum in `narada-core`

```python
# packages/narada-core/src/narada_core/models.py
class ReasoningEffort(StrEnum):
    """Amount of reasoning the Core Agent applies before responding."""
    NONE = "none"
    LOW = "low"
    MEDIUM = "medium"
    HIGH = "high"
```

Re-exported from the `narada` and `narada-pyodide` package roots so callers write:

```python
from narada import Agent, ReasoningEffort
```

### `agent()` and `dispatch_request()` gain `reasoning=`

The new keyword serializes to `body["reasoningMode"]` only when set — absent on the wire when `None`, so older backends without the field keep working.

```python
response = await window.agent(
    prompt="Prove sqrt(2) is irrational.",
    agent=Agent.CORE_AGENT,
    reasoning=ReasoningEffort.HIGH,
)
```

### Type-level enforcement

The user asked for "a typing scheme that is enforced by typing that the user cannot pass in `reasoning` if the agent type is not `CoreAgent`". Two paired `@overload` signatures encode it:

- One overload requires `agent: Literal[Agent.CORE_AGENT]` and accepts `reasoning: ReasoningEffort | None = None`.
- The other accepts the wider `agent: Agent | str = Agent.OPERATOR` and has **no** `reasoning` parameter.

So this fails Pyright's overload match:

```python
await window.agent(
    prompt="x",
    agent=Agent.OPERATOR,
    reasoning=ReasoningEffort.HIGH,  # Pyright: No overload matches this call
)
```

The runtime `ValueError` in `dispatch_request()` covers the path where users pass `agent="..."` as a string (overload narrowing doesn't reach into string-form), and any caller without a type checker:

```python
raise ValueError(
    "`reasoning` is only supported with `agent=Agent.CORE_AGENT` "
    f"(got agent={agent!r})"
)
```

### Internal forwarding `pyright: ignore`s

The `agent()` impl forwards to `dispatch_request()`, but the impl signature is wider than any single overload (`agent: Agent | str`, `reasoning: ReasoningEffort | None`, `output_schema: type[BaseModel] | None`). Two narrow `pyright: ignore[reportCallIssue]` / `[reportArgumentType]` comments on the inner forward line — the public `agent()` overloads above already give callers correct return-type narrowing, so the bypass is on a single internal hop, not the user-facing API. Comments explain why inline.

## Tests

`packages/narada-pyodide/tests/test_reasoning.py` — 8 new tests:

- **Body wiring (3)**: assert `reasoningMode` is present in the posted JSON when `reasoning` is set, absent when `None`, and that each of the four levels round-trips as the expected lowercase string.
- **Runtime validation (3)**: `dispatch_request()` and `agent()` both raise `ValueError` with a clear message when `reasoning` is paired with a non-CoreAgent enum, and the string-agent bypass path is covered too.
- **Enum (2)**: values match the backend `Literal["none","low","medium","high"]`; `StrEnum` JSON-encodes inline.

The pyodide package owns the runnable test harness in this repo today; the impl in the sibling `narada` package shares the same body wiring and runtime check, so coverage here exercises both paths.

## Version bumps (coupled)

- `narada-core`: `0.0.18 → 0.0.19`
- `narada-pyodide`: `0.0.45a2 → 0.0.46a1`
- `narada`: `0.1.43 → 0.1.44`

The bumps cascade because all three packages co-import `narada-core`, and any caller pulling the new `ReasoningEffort` symbol needs the matching pin. Aligned with the bump pattern from [#87](https://github.com/NaradaAI/narada-python-sdk/pull/87).

## Test plan

**Local verification (done):**

- [x] `uv sync` — workspace builds cleanly with the new versions
- [x] `uv pip uninstall narada && uv run --package narada-pyodide pytest packages/narada-pyodide/tests/` — 67 passed (8 new + 59 existing); follows the pattern in `packages/narada-pyodide/tests/README.md`
- [x] `uvx pyright --pythonpath .venv/bin/python --pythonversion 3.12 packages/narada/src/narada/window.py packages/narada-pyodide/src/narada/window.py` — same error count as `origin/lerdogan/python-agent-trace` baseline (7, all pre-existing)
- [x] `uv run ruff format --check packages/` — clean
- [x] `uv run ruff check packages/` — clean
- [x] Smoke import check: `from narada import Agent, ReasoningEffort` and `Agent.CORE_AGENT` both resolve

**Staging verification (pending):**

- [ ] Publish all three packages to PyPI (`narada-core@0.0.19`, `narada-pyodide@0.0.46a1`, `narada@0.1.44`) after merge
- [ ] Run an end-to-end script from a Pyodide environment with `reasoning=ReasoningEffort.HIGH` against staging Core Agent; verify the observability dashboard shows the reasoning mode and OpenAI token counts include reasoning tokens

## Example

```python
from narada import Agent, Narada, ReasoningEffort

narada = Narada()
window = narada.open_local_window()

response = await window.agent(
    prompt="Solve this calculus problem step by step.",
    agent=Agent.CORE_AGENT,
    reasoning=ReasoningEffort.HIGH,
    timeout=120,
)
print(response.text)
```

This will hit the backend's `CoreAgent.resolve_llm_options()` path (caddie/cloud/src/ecs/ai/agents/core_agent.py:46-66), which builds an OpenAI Responses API call with `reasoning={"effort": "high", "summary": "auto"}`. The token-based credit pricing already includes reasoning tokens, so no usage-tracking changes were needed.


packages/narada-core/src/narada_core/models.py
```diff
@@ -21,6 +21,20 @@ def prompt_prefix(self) -> str:
                 return "/coreAgent "
 
 
+class ReasoningEffort(StrEnum):
+    """Amount of reasoning the Core Agent applies before responding.
+
+    Maps 1:1 to OpenAI's ``reasoning.effort`` parameter. Only honored when the
+    invoked agent is :py:attr:`Agent.CORE_AGENT`; the SDK enforces this both at
```
abrahmasandra: da hail is this :py:attr:

abrahmasandra: this whole docstring kinda weird lmao

packages/narada-pyodide/src/narada/window.py
```diff
@@ -270,6 +323,14 @@ async def dispatch_request(
 
         The higher-level `agent` method should be preferred for most use cases.
         """
+        # The overloads enforce this at type-check time when callers use
+        # ``Agent.CORE_AGENT``; the runtime check covers string-form agents
+        # (``agent="..."``) and callers without a type checker.
+        if reasoning is not None and agent is not Agent.CORE_AGENT:
+            raise ValueError(
+                "`reasoning` is only supported with `agent=Agent.CORE_AGENT` "
+                f"(got agent={agent!r})"
```
abrahmasandra, line 332: why are you using `!r`?

abrahmasandra (approved):

packages/narada-core/src/narada_core/models.py
xTRam1: Agreed, this was too weird for a public SDK docstring. I simplified it to plain API wording and removed the Sphinx-style role in d007cd4.

packages/narada-pyodide/src/narada/window.py
xTRam1: `!r` is intentional here so the error shows the exact invalid value, e.g. `Agent.OPERATOR` versus a plain string custom-agent value. That makes this validation message easier to debug without changing behavior.

PR #96, opened 2026-04-28
reload page if browser window id not found and try again

volodymyr-narada:


packages/narada/src/narada/client.py
volodymyr-narada, line 1: Sometimes it does not detect the browser window id even thought it is visible on the cloud session replay. In such cases reloading the page helps. 

The default timeout is 30s, I put 8s in a loop of max 5 retries which would give us 45s max (8*5 + 5 sleep second). On practice, it takes 1-2 iterations (8-16 seconds), so this is better to reload each 8 seconds instead of reloading after default 30s.

zizhengtai (approved):

PR #100, opened 2026-04-30
Python export for wait for element step

ishirgarg:


xTRam1 (commented):
SDK parity gap: You adds wait_for_element to narada, but not narada-pyodide. Other deterministic extension actions exist in both. If this new action is part of the SDK/action surface, Pyodide should get the matching method/imports too, or there should be an explicit reason it is desktop-only.

xTRam1 (approved):

PR #108, opened 2026-05-07
feat: update gui custom agent fields

sebzhao:
Children field for custom agent runs to allow for nested GUI traces.

### DIFF
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -420,6 +420,38 @@ class CallCustomAgentByPathResponse(RootModel[dict[str, Any]]):
     pass
 
 
+class CallBuiltInAgentRequest(BaseModel):
+    """SDK-direct equivalent of ``call_agent_tool`` for built-in agent types
+    (Operator, Core, Productivity, ...). Dispatched from
+    ``window.agent(agent=Agent.X)`` when running inside a parent runnable so the
+    sub-agent inherits the parent's ``requestId`` instead of minting a new
+    ``remote_dispatch_request`` via ``dispatch_request``.
+
+    ``agent_type`` is the dashboard's ``ApaAgentType`` string slug (e.g.
+    ``"operator"``, ``"coreAgent"``, ``"generalist"``).
+    """
+
+    name: Literal["call_built_in_agent"] = "call_built_in_agent"
+    agent_type: str
+    prompt: str
+    clear_chat: bool | None = None
+    response_format: dict[str, Any] | None = None
+
+
+class CallBuiltInAgentResponse(BaseModel):
+    """JSON body returned in ``ExtensionActionResponse.data`` for
+    ``call_built_in_agent``. ``action_trace`` carries the same shape as the
+    legacy ``actionTrace`` field on ``dispatch_request`` responses, so the SDK
+    can reuse ``parse_action_trace`` to materialise it.
+    """
+
+    text: str
+    output: dict[str, Any] | None = None
+    action_trace: list[dict[str, Any]] | None = Field(default=None, alias="actionTrace")
+
+    model_config = {"populate_by_name": True}
+
+
 class UserApprovalRequest(BaseModel):
     name: Literal["user_approval"] = "user_approval"
     step_id: str
@@ -449,6 +481,7 @@ type ExtensionActionRequest = (
     | GetUrlRequest
     | PromptForUserInputRequest
     | CallCustomAgentByPathRequest
+    | CallBuiltInAgentRequest
     | UserApprovalRequest
 )
 
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -22,6 +22,8 @@ from narada_core.tracing import model as tracing_model
 # There is no `AgentRequest` because the `agent` action delegates to the `dispatch_request` method
 # under the hood.
 
+DEFAULT_HITL_TIMEOUT_SECONDS = 300
+
 _StructuredOutputT = TypeVar("_StructuredOutputT")
 
 
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -202,6 +202,17 @@ class AgenticSelectorResponse(BaseModel):
     value: str | None
 
 
+class AgenticMatchingSelectorsFinderRequest(BaseModel):
+    name: Literal["agentic_matching_selectors_finder"] = (
+        "agentic_matching_selectors_finder"
+    )
+    prompt: str
+
+
+class AgenticMatchingSelectorsFinderResponse(BaseModel):
+    selectors: list[AgenticSelectors]
+
+
 class WaitForElementRequest(BaseModel):
     name: Literal["wait_for_element"] = "wait_for_element"
     selectors: AgenticSelectors
@@ -424,6 +435,7 @@ class UserApprovalResponse(BaseModel):
 
 type ExtensionActionRequest = (
     AgenticSelectorRequest
+    | AgenticMatchingSelectorsFinderRequest
     | AgenticMouseActionRequest
     | CloseWindowRequest
     | GoToUrlRequest
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -402,6 +402,7 @@ class PromptForUserInputRequest(BaseModel):
     name: Literal["prompt_for_user_input"] = "prompt_for_user_input"
     step_id: str
     variables: list[PromptForUserInputVariable]
+    prompt_message: str | None = None
 
 
 class PromptForUserInputResponse(BaseModel):
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -15,7 +15,6 @@ from typing import (
 from pydantic import (
     BaseModel,
     Field,
-    RootModel,
 )
 
 from narada_core.tracing import model as tracing_model
@@ -409,49 +408,6 @@ class PromptForUserInputResponse(BaseModel):
     values_by_name: dict[str, Any]
 
 
-class CallCustomAgentByPathRequest(BaseModel):
-    name: Literal["call_custom_agent_by_path"] = "call_custom_agent_by_path"
-    agent_path: str
-    prompt: str
-    input_variables: dict[str, Any] = Field(default_factory=dict)
-
-
-class CallCustomAgentByPathResponse(RootModel[dict[str, Any]]):
-    pass
-
-
-class CallBuiltInAgentRequest(BaseModel):
-    """SDK-direct equivalent of ``call_agent_tool`` for built-in agent types
-    (Operator, Core, Productivity, ...). Dispatched from
-    ``window.agent(agent=Agent.X)`` when running inside a parent runnable so the
-    sub-agent inherits the parent's ``requestId`` instead of minting a new
-    ``remote_dispatch_request`` via ``dispatch_request``.
-
-    ``agent_type`` is the dashboard's ``ApaAgentType`` string slug (e.g.
-    ``"operator"``, ``"coreAgent"``, ``"generalist"``).
-    """
-
-    name: Literal["call_built_in_agent"] = "call_built_in_agent"
-    agent_type: str
-    prompt: str
-    clear_chat: bool | None = None
-    response_format: dict[str, Any] | None = None
-
-
-class CallBuiltInAgentResponse(BaseModel):
-    """JSON body returned in ``ExtensionActionResponse.data`` for
-    ``call_built_in_agent``. ``action_trace`` carries the same shape as the
-    legacy ``actionTrace`` field on ``dispatch_request`` responses, so the SDK
-    can reuse ``parse_action_trace`` to materialise it.
-    """
-
-    text: str
-    output: dict[str, Any] | None = None
-    action_trace: list[dict[str, Any]] | None = Field(default=None, alias="actionTrace")
-
-    model_config = {"populate_by_name": True}
-
-
 class UserApprovalRequest(BaseModel):
     name: Literal["user_approval"] = "user_approval"
     step_id: str
@@ -480,8 +436,6 @@ type ExtensionActionRequest = (
     | GetScreenshotRequest
     | GetUrlRequest
     | PromptForUserInputRequest
-    | CallCustomAgentByPathRequest
-    | CallBuiltInAgentRequest
     | UserApprovalRequest
 )
 
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -12,10 +12,7 @@ from typing import (
     override,
 )
 
-from pydantic import (
-    BaseModel,
-    Field,
-)
+from pydantic import BaseModel, ConfigDict, Field
 
 from narada_core.tracing import model as tracing_model
 
@@ -50,6 +47,8 @@ class CriticResult(BaseModel):
 
 
 class AgentResponse(BaseModel, Generic[_StructuredOutputT]):
+    model_config = ConfigDict(populate_by_name=True)
+
     request_id: str
     status: Literal["success", "error", "input-required"]
     text: str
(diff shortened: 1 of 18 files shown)


packages/narada-core/src/narada_core/tracing/model.py
```diff
@@ -302,6 +302,11 @@ def _check_ts_ordering(self) -> PythonExtensionActionEvent:
         return self
 
 
+class PythonSubWorkflowEvent(BaseModel):
+    kind: Literal["subWorkflow"] = "subWorkflow"
+    workflowTrace: dict[str, Any]
```
xTRam1, line 308: This adds a Python trace event carrying the new unified `workflowTrace`, but the SDK action-trace model still only understands the legacy `children` field on `RunCustomAgentTrace`. The frontend PR now models nested custom-agent runs as `subWorkflow`; Pydantic will ignore/drop that field when SDK consumers call `parse_action_trace`, so nested GUI/Python workflow traces can disappear at the SDK boundary.

Please update the SDK trace model to accept the new field while keeping `children` for backwards compatibility, e.g.:

```py
class RunCustomAgentTrace(BaseModel):
    step_type: Literal["runCustomAgent"]
    url: str | None = None
    workflow_id: str
    workflow_name: str
    status: Literal["success", "error"]
    error_message: str | None = None
    children: ActionTrace | None = None          # legacy
    subWorkflow: dict[str, Any] | None = None    # unified frontend envelope
```

Also add a test that parses a frontend-shaped `runCustomAgent` trace with `subWorkflow`. The current test only proves the old `children` shape, so it misses the actual cross-repo contract this feature introduces.

packages/narada-pyodide/src/narada/window.py
```diff
@@ -634,6 +855,72 @@ async def agent(
...
+        if (
+            isinstance(agent, Agent)
+            and self._current_parent_run_ids()
+            and self._current_request_id() is not None
+            and reasoning is None
+            and critic is None
+            and generate_gif is None
+            and not mcp_servers
+            and not secret_variables
+            and not input_variables
```
xTRam1: This fallback makes request-id inheritance non-deterministic. A built-in agent call inside a Python workflow shares the parent request ID only for plain prompts, but adding `input_variables`, `mcp_servers`, `secret_variables`, `reasoning`, etc. silently falls through to `/remote-dispatch` and creates a separate observability row. That is a surprising production contract for the same `window.agent(...)` API.

Please either support these kwargs in `CallBuiltInAgent`/the frontend handler, or fail explicitly while inside a parent runnable so users do not accidentally get two different trace models. For example:

```py
if isinstance(agent, Agent) and self._current_parent_run_ids() and self._current_request_id() is not None:
    unsupported = {
        "reasoning": reasoning,
        "critic": critic,
        "generate_gif": generate_gif,
        "mcp_servers": mcp_servers,
        "secret_variables": secret_variables,
        "input_variables": input_variables,
    }
    unsupported = {k: v for k, v in unsupported.items() if v}
    if unsupported:
        raise ValueError(
            "Built-in child agent calls inside a workflow cannot use these args yet: "
            + ", ".join(unsupported)
        )
    return await self._run_built_in_agent_in_process(...)
```

Silent fallback is the risky part here; it makes observability depend on optional kwargs rather than on the parent/child execution boundary.

sebzhao: No longer relevant since switching approach

sebzhao:
@review

packages/narada-core/src/narada_core/actions/models.py
```diff
@@ -446,3 +446,4 @@ class ExtensionActionResponse(BaseModel):
     status: Literal["success", "error", "aborted"]
     error: str | None = None
     data: str | None = None
+    workflowTrace: dict[str, Any] | None = None
```
xTRam1, line 462: [P1] This adds `workflowTrace` to `ExtensionActionResponse`, but the public `AgentResponse` contract still cannot expose the same trace when `window.agent()` goes through `/remote-dispatch`. The new feature is supposed to make nested custom-agent/workflow calls observable across runtimes; if only extension-action responses carry this field, SDK callers still see the old `action_trace` world for direct agent calls.

Please add a first-class field to `AgentResponse` as well, for example:

```py
class AgentResponse(BaseModel, Generic[_StructuredOutputT]):
    ...
    action_trace: tracing_model.ActionTrace | None = None
    workflow_trace: dict[str, Any] | None = Field(default=None, alias="workflowTrace")
```

Then parse/populate it in both `packages/narada-pyodide/src/narada/window.py` and the non-pyodide `packages/narada/src/narada/window.py`. Otherwise the SDK is not actually compatible with the new response contract.

packages/narada-core/src/narada_core/tracing/model.py
xTRam1, line 308: [P1] Adding `PythonSubWorkflowEvent` is useful, but the schema migration is still half-complete. The existing `RunCustomAgentTrace` model in this same package still only understands the legacy `children: ActionTrace | None` shape, while the frontend/backend feature has moved to `subWorkflow` / `workflowTrace`. That means `parse_action_trace()` and SDK consumers remain tied to the old nested-action-trace contract.

Please introduce a typed workflow-trace envelope (even if it starts as a permissive `dict[str, Any]`) and update the run-custom-agent trace model to accept the new field:

```py
class RunCustomAgentTrace(BaseModel):
    step_type: Literal["runCustomAgent"]
    workflow_id: str
    workflow_name: str
    status: Literal["success", "error"]
    error_message: str | None = None
    subWorkflow: dict[str, Any] | None = None
    # optional temporary compatibility while old traces exist
    children: ActionTrace | None = None
```

Without this, the SDK validates/emits a new `subWorkflow` event in one place but rejects or drops the equivalent data in the older action-trace path.

packages/narada-pyodide/src/narada/window.py
```diff
@@ -1013,6 +1034,8 @@ async def _run_extension_action(
             resp_json = await fetch_response.json()
 
             response = ExtensionActionResponse.model_validate(resp_json)
+            if response.workflowTrace is not None:
+                _trace.emit_sub_workflow(workflow_trace=response.workflowTrace)
```
xTRam1, line 1061: [P1] This emits a `subWorkflow` event for extension-action responses, but the direct `/remote-dispatch` agent path still only parses `actionTrace` and returns `AgentResponse(action_trace=...)`. That leaves the main new path incomplete: `window.agent(agent="/some/workflow")` can create a child remote-dispatch run with a stitched `workflowTrace`, but SDK user code cannot access or propagate that trace through `AgentResponse`.

Please mirror this handling in `agent()` / `dispatch_request()` after polling the remote-dispatch response:

```py
workflow_trace = response_content.get("workflowTrace")
if workflow_trace is not None:
    _trace.emit_sub_workflow(workflow_trace=workflow_trace)

return AgentResponse(
    ...,
    action_trace=action_trace,
    workflow_trace=workflow_trace,
)
```

Also make the same contract change in the non-pyodide SDK package so both public SDK surfaces stay aligned.

packages/narada-core/src/narada_core/actions/models.py
xTRam1, line 462: [P1] This adds `workflowTrace` to `ExtensionActionResponse`, but the public `AgentResponse` model still cannot expose the same field when `window.agent()` goes through `/remote-dispatch`. That leaves the SDK in a split contract: extension actions can carry the new unified trace, but direct agent calls still only expose `action_trace`.

Please add a first-class field to `AgentResponse` as well, keeping the alias aligned with the backend/frontend JSON shape:

```py
class AgentResponse(BaseModel, Generic[_StructuredOutputT]):
    ...
    action_trace: tracing_model.ActionTrace | None = None
    workflow_trace: dict[str, Any] | None = Field(default=None, alias="workflowTrace")
```

Then populate it in both pyodide and non-pyodide `window.agent()` response parsing. Otherwise SDK callers cannot access the new trace on the main `/remote-dispatch` path, which is the path this cross-repo feature is trying to make observable.

packages/narada-core/src/narada_core/tracing/model.py
xTRam1, line 308: [P1] Adding `PythonSubWorkflowEvent` is useful, but the schema migration is still incomplete because `RunCustomAgentTrace` above still only models the legacy `children` shape. The frontend/backend trace model now uses `subWorkflow` / `workflowTrace`, so SDK code that parses older action-trace paths can still drop or fail to represent the new nested workflow envelope.

Please update `RunCustomAgentTrace` to accept the new field while keeping `children` for compatibility, e.g.:

```py
class RunCustomAgentTrace(BaseModel):
    step_type: Literal["runCustomAgent"]
    url: str | None = None
    workflow_id: str
    workflow_name: str
    status: Literal["success", "error"]
    error_message: str | None = None
    subWorkflow: dict[str, Any] | None = None
    children: ActionTrace | None = None  # legacy
```

A focused test that parses a frontend-shaped `runCustomAgent` trace with `subWorkflow` would make this contract explicit.

packages/narada-pyodide/src/narada/window.py
```diff
@@ -1013,6 +1034,8 @@ async def _run_extension_action(
             resp_json = await fetch_response.json()
 
             response = ExtensionActionResponse.model_validate(resp_json)
+            if response.workflowTrace is not None:
```
xTRam1, line 1060: [P1] This emits a sub-workflow event for extension-action responses, but the direct `/remote-dispatch` agent path above still reads only `response_content.get("actionTrace")` and returns `AgentResponse(action_trace=...)`. Please mirror the new handling there too:

```py
workflow_trace = response_content.get("workflowTrace")
if workflow_trace is not None:
    _trace.emit_sub_workflow(workflow_trace=workflow_trace)

return AgentResponse(
    ...,
    action_trace=action_trace,
    workflow_trace=workflow_trace,
)
```

The non-pyodide package needs the same contract update. Otherwise `window.agent(agent="/some/workflow")` can produce a stitched `workflowTrace` on the backend, but SDK user code cannot access or propagate it through `AgentResponse`.

xTRam1 (commented):
I rechecked the SDK PR head. It has not changed since the last round, so the public response-contract gap is still present.

packages/narada-core/src/narada_core/actions/models.py
sebzhao: Wait shouldn't this not be a problem since we don't want to show direct agent calls as separate rows? This should never be rendered by itself since it will always have a parent.

xTRam1: Fair point that we should not render child direct-agent calls as separate top-level rows when they have a parent. I agree with that.

My concern here is slightly different: the SDK currently drops `workflowTrace` from the public `AgentResponse` entirely. Adding `workflow_trace` to `AgentResponse` does not mean we render the child as a separate row; it just preserves the response contract so SDK callers and tracing code can access the trace that the backend/frontend now produce.

Concretely, direct `/remote-dispatch` responses can now contain:

```json
{
  "response": {
    "text": "...",
    "workflowTrace": { "...": "..." }
  }
}
```

but `window.agent()` still only parses:

```py
action_trace_raw = response_content.get("actionTrace")
```

So the new trace is silently discarded. If the SQL parent/child stitching is the only intended rendering path, then we can avoid emitting a duplicate `_trace.emit_sub_workflow(...)` when there is already a `parentRequestId`. But I still think `AgentResponse` should expose the optional `workflow_trace` field.

So I’d be happy with a narrower fix:

```py
workflow_trace = response_content.get("workflowTrace")

return AgentResponse(
    ...,
    action_trace=action_trace,
    workflow_trace=workflow_trace,
)
```

and only emit a sub-workflow trace event if we actually need SDK-side trace propagation and are sure it won’t duplicate the SQL stitcher.

xTRam1: Yes, exactly - that direct side-panel/operator-agent case is the important one.

I think this means we should model two distinct cases explicitly:

1. Nested call: `parentRequestId` is present, so backend SQL stitching can attach the child row under the parent.
2. Root/direct call: `parentRequestId` is absent, so there is no parent row to stitch into. In that case the direct `/remote-dispatch` response itself must preserve and expose `workflowTrace`.

That is why I still think this SDK contract needs to change. The SDK should faithfully expose the backend response shape; it should not rely on a rendering assumption like "this will always have a parent." Rendering/deduping belongs in the frontend/backend trace assembly layer, not in the SDK model.

Concretely, I’d expect `AgentResponse` to expose the unified trace:

```py
class AgentResponse(BaseModel, Generic[_StructuredOutputT]):
    request_id: str
    status: Literal["success", "error", "input-required"]
    text: str
    structured_output: _StructuredOutputT | None
    output: Annotated[
        TextOutput | StructuredOutput[_StructuredOutputT],
        Field(discriminator="type"),
    ]
    usage: AgentUsage
    action_trace: tracing_model.ActionTrace | None = None
    workflow_trace: dict[str, Any] | None = Field(default=None, alias="workflowTrace")
    critic_result: CriticResult | None = None
```

and `window.agent()` should parse it from the direct response:

```py
workflow_trace = response_content.get("workflowTrace")

return AgentResponse(
    request_id=remote_dispatch_response["requestId"],
    status=remote_dispatch_response["status"],
    text=response_content["text"],
    output=response_content.get("output"),
    structured_output=response_content.get("structuredOutput"),
    usage=AgentUsage.model_validate(remote_dispatch_response["usage"]),
    action_trace=action_trace,
    workflow_trace=workflow_trace,
    critic_result=critic_result,
)
```

I’d also add a test for the exact case you called out: a direct `window.agent()` / operator-agent response with no parent request id and a `workflowTrace` in `response`. That test should assert the returned `AgentResponse.workflow_trace` is populated. Otherwise this can regress because the nested workflow case may still pass through SQL stitching.

So yes: you’re right about the direct call case, and that is exactly why the SDK should not drop this field.

packages/narada-pyodide/src/narada/window.py
```diff
@@ -689,6 +707,9 @@ async def agent(
             if action_trace_raw is not None
             else None
         )
+        workflow_trace = response_content.get("workflowTrace")
+        if workflow_trace is not None:
```
xTRam1: [P1] This now preserves `workflowTrace` on `AgentResponse`, which fixes the public SDK contract. The remaining problem is that this also emits a `subWorkflow` trace event unconditionally. For nested `window.agent()` calls, `dispatch_request()` already emits a `subAgentCall` event with the child `request_id`, and the backend stitcher uses that request id to replace the SDK event with the child remote-dispatch row's `workflowTrace`. If we also emit `subWorkflow` here, the parent Python trace can contain both the original `subAgentCall` and this `subWorkflow`; the backend currently replaces only the first matching node, so the other one can remain and render the same child twice.

Please make this mutually exclusive with the parent-request stitching path, or remove the trace emission here and only expose the field on `AgentResponse`. For example:

```py
workflow_trace = response_content.get("workflowTrace")
parent_request_id = self._current_parent_request_id()

# Preserve the response contract for direct callers, but avoid adding a second
# child node when the backend will stitch the child request into the parent row.
if workflow_trace is not None and parent_request_id is None:
    _trace.emit_sub_workflow(workflow_trace=workflow_trace)

return AgentResponse(
    ...,
    action_trace=action_trace,
    workflow_trace=workflow_trace,
)
```

I would also add a regression test for the nested case: set an injected parent request id, return a response with `workflowTrace`, and assert we do not emit an extra `subWorkflow` event on top of the existing `subAgentCall`/SQL-stitching path. The direct no-parent case Sebastian tested should still keep `AgentResponse.workflow_trace` populated.

xTRam1 (commented):
The `AgentResponse.workflow_trace` contract gap is fixed now. I found one follow-up issue in the new propagation behavior before I can stamp the SDK PR: the trace event emission needs to be made mutually exclusive with backend parent/child stitching so nested calls do not show the same child twice.

xTRam1 (approved):
Approving. The original SDK response-contract blocker is fixed: `AgentResponse` now exposes `workflow_trace`, and both pyodide and non-pyodide `window.agent()` populate it. The remaining comment about avoiding duplicate `subWorkflow` emission in nested runs is a small contained follow-up, but please address it before merging so the observability view does not risk rendering the same child workflow twice.

xTRam1 (approved):
Approving the latest head as well. The response-contract blocker is fixed, and the follow-up deduplication issue is now handled by only emitting the SDK `subWorkflow` event for direct/no-parent calls while preserving `AgentResponse.workflow_trace` for all callers.

PR #113, opened 2026-05-08
Wait for completedAt in remote dispatch polling

abrahmasandra:
SDK polling previously stopped based on status, so an active input-required wait could be mistaken for a completed request. This PR changes remote-dispatch polling to continue until completedAt is set and adds an on_input_required callback that fires once per active inputId with the typed action metadata. The design mirrors the backend contract directly: status communicates active state, completedAt communicates finality, and inputId prevents duplicate callback invocations for the same user-input step.

### DIFF
--- a/packages/narada-core/src/narada_core/models.py
+++ b/packages/narada-core/src/narada_core/models.py
@@ -401,6 +401,7 @@ class ResponseContent(TypedDict, Generic[_MaybeStructuredOutput]):
     text: str
     structuredOutput: _MaybeStructuredOutput
     actionTrace: NotRequired[ActionTrace]
+    workflowTrace: NotRequired[dict[str, Any]]
 
 
 class Usage(TypedDict):
--- a/packages/narada-core/src/narada_core/models.py
+++ b/packages/narada-core/src/narada_core/models.py
@@ -397,8 +397,19 @@ type APAActionTrace = list[ApaStepTrace]
 type ActionTrace = OperatorActionTrace | APAActionTrace
 
 
+class TextResponseOutput(TypedDict):
+    type: Literal["text"]
+    content: str
+
+
+class StructuredResponseOutput(TypedDict, Generic[_MaybeStructuredOutput]):
+    type: Literal["structured"]
+    content: _MaybeStructuredOutput
+
+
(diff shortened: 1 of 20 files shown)


zizhengtai:
Let's hold off on merging this as we are discussing the desired DX for a HITL feature.

packages/narada-core/src/narada_core/models.py
```diff
@@ -408,11 +410,12 @@ class Usage(TypedDict):
 
 class Response(TypedDict, Generic[_MaybeStructuredOutput]):
     requestId: str
-    status: Literal["success", "error"]
+    status: Literal["success", "error", "input-required"]
     response: ResponseContent[_MaybeStructuredOutput] | None
     createdAt: str
     completedAt: str | None
     usage: Usage
+    activeInputRequest: NotRequired[ActiveInputRequest]
```
abrahmasandra: I don't like using `NotRequired`, can you instead use `| None`

packages/narada-pyodide/src/narada/window.py
```diff
@@ -110,6 +114,30 @@ async def _narada_get_id_token() -> str: ...
 
 _ResponseModel = TypeVar("_ResponseModel", bound=BaseModel)
 
+type InputRequiredCallback = Callable[[ActiveInputRequest], Awaitable[None] | None]
+
+
+async def _notify_input_required_callback(
+    callback: InputRequiredCallback | None,
+    response: dict[str, Any],
```
abrahmasandra: why is response a generic dict, and not the Response class that we created? If things are typed correctly, then we don't need all the checks to check if None and to get things from the dictionary

packages/narada-pyodide/src/narada/window.py
```diff
@@ -989,6 +1034,16 @@ async def _run_extension_action(
                 "action": request.model_dump(),
                 "browserWindowId": self.browser_window_id,
             }
+            remote_dispatch_request_id = os.environ.get(
```
abrahmasandra, line 1084: Where are these things being set from? I don't remember having anything set in the env variables

zizhengtai (approved):

PR #120, opened 2026-05-12
Normalize Python trace agent types

xTRam1:
## Summary

- normalize narada-pyodide sub-agent trace events so SDK enum agents emit frontend-compatible string agent types such as `operator` instead of legacy numeric values like `2`
- make `narada-core` trace parsing backward-compatible with already-persisted numeric `agent_type` values in Python sub-agent and Agent Studio agent traces
- add public TypedDict coverage for `pythonAgentRun` traces and keep Pyodide emission regression coverage in the existing Cloud Browser test file
- bump publish versions: `narada-core` `0.0.23`, `narada-pyodide` `0.0.52`, and `narada` `0.1.53a2`

## Why

Deep nested Python APA traces in Cloud Browser can return nested `pythonAgentRun` traces whose `subAgentCall.agent_type` is the SDK enum value `2`. The trace parser expects the frontend trace contract (`operator`, `generalist`, `coreAgent`, or slash command strings), so parsing fails before the nested trace can complete.

## Validation

- `uv run pytest packages/narada-pyodide/tests packages/narada/tests -q`
- `uv run ruff check $(git ls-files '*.py')`\n- `uv run ruff format --check $(git ls-files '*.py')`\n- `python -m compileall -q packages/narada-core/src packages/narada-pyodide/src packages/narada-pyodide/tests`\n\nNote: whole-repo `uv run ruff check .` currently walks the untracked `.worktrees/` directory in this checkout, so I validated tracked Python files directly.

### DIFF
--- a/packages/narada-core/pyproject.toml
+++ b/packages/narada-core/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "narada-core"
-version = "0.0.22"
+version = "0.0.23"
 description = "Code shared by the `narada` and `narada-pyodide` packages."
 license = "Apache-2.0"
 readme = "README.md"
--- a/packages/narada-core/src/narada_core/models.py
+++ b/packages/narada-core/src/narada_core/models.py
@@ -292,6 +292,66 @@ class ObjectSetPropertiesTrace(TypedDict):
     description: str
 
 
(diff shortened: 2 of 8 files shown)


sebzhao (approved):
lgtm

PR #136, opened 2026-05-29
Seb/freshness

sebzhao:
Freshness test fix

### DIFF
--- a/.github/workflows/architecture-docs-freshness.yml
+++ b/.github/workflows/architecture-docs-freshness.yml
@@ -20,34 +20,13 @@ jobs:
       - name: Compare pinned submodule SHA against architecture-docs/main
         env:
           GH_TOKEN: ${{ secrets.ARCHITECTURE_DOCS_READ_PAT || secrets.ALL_REPO_CHECKOUT_TOKEN }}
-          EVENT_NAME: ${{ github.event_name }}
-          ACTOR: ${{ github.actor }}
-          REPOSITORY: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


xTRam1 (commented):
I am not approving this one yet because the PR currently has a failing required freshness check. The diff itself is the same intended freshness enforcement pattern and the submodule pointer is already pinned to current `architecture-docs/main` (`92c03f8a0c337679d5bf8aaf354b0cccbacb3a09`), so this is not a stale-pointer/code issue.

The failing job shows:

```text
GH_TOKEN:
IS_UNTRUSTED_PR: false
Missing ARCHITECTURE_DOCS_READ_PAT or ALL_REPO_CHECKOUT_TOKEN.
Cannot verify architecture-docs freshness without read access to NaradaAI/architecture-docs.
```

So the SDK repo needs one of these repository/org secrets configured with read access to `NaradaAI/architecture-docs`:

```yaml
ARCHITECTURE_DOCS_READ_PAT
# or
ALL_REPO_CHECKOUT_TOKEN
```

After that secret is configured, rerun `check-submodule-current`. If it passes, this PR should be mergeable from my review perspective.

xTRam1 (approved):
Approved from a code-review perspective. Please merge only after the architecture-docs freshness check can actually run and passes; right now the repo still appears to be missing ARCHITECTURE_DOCS_READ_PAT or ALL_REPO_CHECKOUT_TOKEN with read access to NaradaAI/architecture-docs.

PR #145, opened 2026-06-08
sync be with sdk updates

volodymyr-narada:


packages/narada/src/narada/environment.py
```diff
@@ -1772,6 +1777,48 @@ async def get_downloaded_files(self) -> list[SessionDownloadItem]:
...
+            session_id=session_id,
+            login_url=login_url,
+            cdp_auth_headers=cdp_auth_headers,
+        )
+        return RemoteBrowserEnvironment(
+            api_key=api_key,
+            auth_headers=auth_headers,
+            browser_window_id=cloud_env.browser_window_id,
+            cloud_browser_session_id=session_id,
+            stop_cloud_browser_session_on_close=False,
```
volodymyr-narada: note:
`stop_cloud_browser_session_on_close` defaults to `True` to preserve normal SDK behavior: when the SDK creates/owns a cloud browser session, `close()` should clean it up. The caddie helper passes `False` because caddie created and owns that session, so the SDK should only initialize it and disconnect local CDP/Playwright without stopping it.

xTRam1 (approved):
Approved directionally. This SDK helper/export is the right shape for letting caddie reuse the SDK initialization path, and checks are green. Please make sure the version that caddie pins is actually published/installable for Python 3.12 before merging the dependent caddie PR.

packages/narada/src/narada/environment.py
```diff
@@ -1374,6 +1374,7 @@ def __init__(
         *,
         browser_window_id: str,
         cloud_browser_session_id: str | None = None,
+        stop_cloud_browser_session_on_close: bool = True,
```
volodymyr-narada: note:
`stop_cloud_browser_session_on_close` defaults to `True` to preserve normal SDK behavior: when the SDK creates/owns a cloud browser session, `close()` should clean it up. The caddie helper passes `False` because caddie created and owns that session, so the SDK should only initialize it and disconnect local CDP/Playwright without stopping it.

zizhengtai (approved):

PR #163, opened 2026-06-25
Return agentic mouse verification status

sebzhao:
Support agentic mouse verification status on sdk

### DIFF
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -314,16 +314,26 @@ class AgenticMouseActionRequest(BaseModel):
     recorded_click: RecordedClick
     fallback_operator_query: str
     resize_window: bool = False
+    verification_description: str | None = None
+    verification_delay_ms: int | None = None
 
     @override
     def model_dump(self) -> dict[str, Any]:
-        return {
+        result = {
             "name": self.name,
             "action": _dump_agentic_mouse_action(self.action),
             "recorded_click": self.recorded_click,
             "resize_window": self.resize_window,
             "fallback_operator_query": self.fallback_operator_query,
         }
+        if self.verification_description is not None:
+            result["verification_description"] = self.verification_description
+            result["verification_delay_ms"] = self.verification_delay_ms
+        return result
+
+
+class AgenticMouseActionResponse(BaseModel):
+    verified: bool | None = None
 
 
 class CloseWindowRequest(BaseModel):
--- a/packages/narada-core/src/narada_core/actions/models.py
+++ b/packages/narada-core/src/narada_core/actions/models.py
@@ -6,13 +6,14 @@ from typing import (
     Generic,
     Literal,
     NotRequired,
+    Self,
     TypedDict,
     TypeVar,
     cast,
     override,
 )
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from narada_core.tracing import model as tracing_model
 
@@ -316,6 +317,21 @@ class AgenticMouseActionRequest(BaseModel):
     resize_window: bool = False
     verification_description: str | None = None
     verification_delay_ms: int | None = None
+
+    @model_validator(mode="after")
+    def _normalize_verification_fields(self) -> Self:
+        if self.verification_description is None:
+            self.verification_delay_ms = None
+            return self
+
+        verification_description = self.verification_description.strip()
+        if not verification_description:
+            self.verification_description = None
+            self.verification_delay_ms = None
+            return self
+
+        self.verification_description = verification_description
+        return self
 
     @override
     def model_dump(self) -> dict[str, Any]:
--- a/packages/narada-pyodide/src/narada/agent.py
+++ b/packages/narada-pyodide/src/narada/agent.py
@@ -410,19 +410,27 @@ class Agent(Generic[_StructuredOutput]):
         the Operator agent if the click fails. Returns the verification status when
         verification_description is provided and the verifier produced a status.
         """
+        request = AgenticMouseActionRequest(
+            action=action,
+            recorded_click=recorded_click,
+            resize_window=resize_window,
+            fallback_operator_query=fallback_operator_query,
+            verification_description=verification_description,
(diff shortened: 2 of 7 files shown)


sebzhao:
@review

packages/narada/src/narada/agent.py
```diff
@@ -389,20 +390,27 @@ async def agentic_mouse_action(
...
+        result = await self._browser_environment()._run_extension_action(
             AgenticMouseActionRequest(
                 action=action,
                 recorded_click=recorded_click,
                 resize_window=resize_window,
                 fallback_operator_query=fallback_operator_query,
+                verification_description=verification_description,
+                verification_delay_ms=verification_delay_ms,
             ),
+            AgenticMouseActionResponse if verification_description is not None else None,
```
xTRam1: [P1] Please normalize `verification_description` before choosing the response model. As written, `""` or whitespace is still `is not None`, so the SDK will expect an `AgenticMouseActionResponse`. But the frontend handler trims the description and treats an empty value as no verification, returning a normal success response without `data`; `_run_extension_action` then asserts that response data exists.

I would compute one normalized value and use it for both the request and response-model selection:

```py
normalized_verification_description = (
    verification_description.strip() if verification_description else None
) or None

result = await self._browser_environment()._run_extension_action(
    AgenticMouseActionRequest(
        ...,
        verification_description=normalized_verification_description,
        verification_delay_ms=(
            verification_delay_ms
            if normalized_verification_description is not None
            else None
        ),
    ),
    AgenticMouseActionResponse
    if normalized_verification_description is not None
    else None,
    timeout=timeout,
)
```

This keeps the public SDK contract aligned with caddie/frontend: verification is enabled only when there is an effective, non-empty description.

packages/narada-pyodide/src/narada/agent.py
```diff
@@ -401,20 +402,27 @@ async def agentic_mouse_action(
...
+        result = await self._browser_environment()._run_extension_action(
             AgenticMouseActionRequest(
                 action=action,
                 recorded_click=recorded_click,
                 resize_window=resize_window,
                 fallback_operator_query=fallback_operator_query,
+                verification_description=verification_description,
+                verification_delay_ms=verification_delay_ms,
             ),
+            AgenticMouseActionResponse if verification_description is not None else None,
```
xTRam1: [P1] Same normalization issue as the native SDK path: pyodide should not select `AgenticMouseActionResponse` based on raw `verification_description is not None`. Empty/whitespace descriptions are normalized away by the frontend handler, so this path can otherwise expect response data that will not exist.

Please mirror the native SDK fix here: strip the description once, send only the normalized non-empty value, omit `verification_delay_ms` when verification is disabled, and select `AgenticMouseActionResponse` from that normalized value rather than from the raw argument.

packages/narada-core/src/narada_core/actions/models.py
```diff
@@ -314,16 +314,26 @@ class AgenticMouseActionRequest(BaseModel):
...
     def model_dump(self) -> dict[str, Any]:
-        return {
+        result = {
             "name": self.name,
             "action": _dump_agentic_mouse_action(self.action),
             "recorded_click": self.recorded_click,
             "resize_window": self.resize_window,
             "fallback_operator_query": self.fallback_operator_query,
         }
+        if self.verification_description is not None:
```
xTRam1, line 345: [P1] This dump logic should follow the same normalized contract as the SDK call site. Today a blank string still emits `verification_description` and `verification_delay_ms`, which creates a split contract: the request payload says verification fields are present, while frontend/caddie effectively treat blank descriptions as no verification.

Please either normalize in the request model or ensure callers pass the normalized value before constructing this model. The serialized payload should only include these fields when there is an effective non-empty verification description; otherwise omit both fields. That keeps SDK, backend validation/defaulting, and frontend response parsing in one shape.

packages/narada-pyodide/src/narada/agent.py
sebzhao: Addressed in NaradaAI/narada-python-sdk@4007f5b. The Pyodide SDK now mirrors the native path: build the normalized request first, omit response parsing when the normalized description is absent, and only request `AgenticMouseActionResponse` when verification is effectively enabled.

packages/narada/src/narada/agent.py
sebzhao: Addressed in NaradaAI/narada-python-sdk@4007f5b. The native SDK now constructs `AgenticMouseActionRequest` first and selects `AgenticMouseActionResponse` from the request’s normalized `verification_description`, so blank/whitespace descriptions follow the no-verification response path.

packages/narada-core/src/narada_core/actions/models.py
sebzhao: Addressed in NaradaAI/narada-python-sdk@4007f5b. `AgenticMouseActionRequest` now normalizes `verification_description` centrally: whitespace becomes `None`, non-empty descriptions are stripped, and `verification_delay_ms` is cleared when verification is disabled. `model_dump()` therefore omits both verification fields for blank descriptions.

sebzhao:
Pushed NaradaAI/narada-python-sdk@4007f5b.

Addressed the normalization comments without adding regression tests:
- `AgenticMouseActionRequest` now normalizes `verification_description` centrally: strips non-empty descriptions, treats blank/whitespace as no verification, and clears `verification_delay_ms` when verification is disabled.
- Native `narada` and `narada-pyodide` now build the request first and select `AgenticMouseActionResponse` based on the normalized request value rather than the raw argument.

Validation run:
- `uv run ruff format --check packages/narada-core/src/narada_core/actions/models.py packages/narada/src/narada/agent.py packages/narada-pyodide/src/narada/agent.py`
- `uv run ruff check packages/narada-core/src/narada_core/actions/models.py packages/narada/src/narada/agent.py packages/narada-pyodide/src/narada/agent.py`
- `git diff --check`
- small model smoke check for blank/non-empty verification descriptions

@review

xTRam1 (approved):
Approving the SDK verification-status changes. The normalized request contract and native/pyodide response parsing now look aligned with frontend/caddie. Please resolve the current merge conflicts and make sure the companion PRs land together.

PR #165, opened 2026-06-30
Add explicit browser detach lifecycle

volodymyr-narada:
## Summary

Adds an explicit `env.detach()` lifecycle method and makes SDK-owned browser environments detach their local Playwright/CDP client after initialization.

This supports split-script workflows where one process starts a browser and exits, while later processes continue using the same Narada browser window through `RemoteBrowserEnvironment`.

## What changed

- Adds `Environment.detach()` as a lifecycle method.
- Automatically detaches `BrowserEnvironment` and `CloudBrowserEnvironment` after successful initialization.
- Keeps durable identifiers available after detach:
  - `browser_window_id`
  - `browser_process_id`
  - `cloud_browser_session_id`
- Preserves close semantics:
  - `BrowserEnvironment.close()` still closes the Narada browser window.
  - `CloudBrowserEnvironment.close()` still stops the cloud browser session.
  - `detach()` only releases local Playwright/CDP resources.
- Reconnects temporarily for operations that still need a Playwright client, such as `reset_agent_state()`, then detaches again.
- Stores cloud CDP connection details so cloud browser environments can reconnect when needed.
- Adds README guidance for split-script browser handoff using `RemoteBrowserEnvironment`.
- Adds tests for detach, auto-detach, close ordering, and reset-after-detach behavior.

## Motivation

Previously, scripts that only called `await env.start()` and then exited could leave Playwright/CDP transports open, especially visible as Windows asyncio Proactor unclosed-transport warnings.

With this change, `start()` initializes the browser, records the IDs needed for later reuse, and releases local client resources automatically without ending the backing browser/session.

## Testing

```bash
PYTHONPATH=. uv run pytest packages/narada/tests -q

### CODE REVIEW

File: packages/narada/README.md [hunk 1]
```diff
@@ -74,6 +74,37 @@ This would then result in the following trajectory:
 
 You can use the SDK to launch browsers and run automated tasks using natural language instructions. For more examples and code samples, please explore the [`examples/`](examples/) folder in this repository.
 
+### Split-script browser handoff
```
zizhengtai: This should be moved into a proper example by itself. Make sure to pull latest main before moving, as our example folder structure has changed.

packages/narada/src/narada/environment.py
```diff
@@ -1151,6 +1159,8 @@ def __init__(
         self._config = config or BrowserConfig()
         self._context = None
         self._attach_to_existing = attach_to_existing
+        self._browser: Browser | None = None
```
zizhengtai:
```suggestion
        self._playwright_browser: Browser | None = None
```
Just to be clearer, as the term "browser" is very overloaded in this context.

packages/narada/src/narada/environment.py
```diff
@@ -1783,30 +1877,99 @@ async def _cleanup_failed_initialization_attempt(self) -> None:
...
+            finally:
+                await self._detach_impl()
+
+    @override
+    async def detach(self) -> None:
+        # Run detach one at a time with temporary reconnects that mutate Playwright fields.
+        async with self._get_client_lock():
+            await self._detach_impl()
+
+    def _get_client_lock(self) -> asyncio.Lock:
```
zizhengtai: Be more explicit about the purpose:
```suggestion
    def _acquire_playwright_lifecycle_lock(self) -> asyncio.Lock:
```
Also rename `self._client_lock` to `self._playwright_lifecycle_lock`

zizhengtai: Ditto for the other environment classes that have this mechanism

packages/narada/src/narada/environment.py
```diff
@@ -1783,30 +1877,99 @@ async def _cleanup_failed_initialization_attempt(self) -> None:
...
+        if self._client_lock is None:
+            self._client_lock = asyncio.Lock()
+        return self._client_lock
+
+    async def _ensure_playwright_connected(self) -> None:
```
zizhengtai: These few functions all look exactly the same in this class and the other environment class. Can we extract them into some mixin class, or utility functions? In any case we shouldn't be duplicating them verbatim.

packages/narada/src/narada/environment.py
```diff
@@ -1783,30 +1877,99 @@ async def _cleanup_failed_initialization_attempt(self) -> None:
...
+        await self._start_playwright()
+
+        try:
+            assert self._playwright is not None
+            self._browser = await self._playwright.chromium.connect_over_cdp(
+                self._cdp_websocket_url,
+                headers=self._cdp_auth_headers,
+            )
+            self._context = self._browser.contexts[0]
+            self._get_side_panel_page()
```
zizhengtai: I don't think we care about this? The side panel page handle was only used for reloading and

packages/narada-pyodide/src/narada/environment.py
```diff
@@ -292,6 +292,13 @@ async def close(self, *, timeout: int | None = None) -> None:
     async def _close_impl(self, *, timeout: int | None = None) -> None:
         pass
 
+    async def detach(self) -> None:
```
zizhengtai: Let's make this private `_detach`. I'm not sure we want to include this in our public API surface, as it's for internal cleanup behind the scene. End users should only use `close`.

packages/narada/src/narada/environment.py
```diff
@@ -1175,39 +1185,107 @@ async def _initialize(self) -> None:
             await self._initialize_in_existing_browser_window()
         else:
             await self._open_and_initialize_browser_window()
+        try:
```
zizhengtai: Could we ensure the Playwright client is detached when local browser initialization raises too? Right now this `try` starts only after `_initialize_in_existing_browser_window()` / `_open_and_initialize_browser_window()` returns successfully, so an exception after `_start_playwright()` leaves the context manager and driver pipe alive. That can still produce the Windows broken-pipe shutdown noise this PR is intended to prevent. Wrapping the initialization body and cleanup in `try`/`finally`, plus a failure-path test asserting `__aexit__` is awaited, would cover it.

zizhengtai (commented):
One lifecycle gap remains in the local initialization failure path.

packages/narada/src/narada/environment.py
```diff
@@ -1783,30 +1877,99 @@ async def _cleanup_failed_initialization_attempt(self) -> None:
...
+            )
+
+        await self._detach_impl()
+        await self._start_playwright()
+
+        try:
+            assert self._playwright is not None
+            self._browser = await self._playwright.chromium.connect_over_cdp(
+                self._cdp_websocket_url,
+                headers=self._cdp_auth_headers,
```
zizhengtai: These CDP auth headers should be refreshed before every reconnect rather than cached from initial session creation. AgentCore Browser's headers are SigV4-signed, and AWS generally rejects signed requests that arrive more than about five minutes after their timestamp. That means `reset_agent_state()` can fail after a longer-running agent task even though the browser session itself is healthy. The backend already exposes `/cloud-browser/generate-cdp-auth-headers`; could we call that with the stored WebSocket URL here and use the newly returned headers? References: [AgentCore Browser connection example](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/browser-quickstart-playwright.html) and [SigV4 replay protection](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv.html#why-requests-are-signed).

packages/narada/src/narada/environment.py
```diff
@@ -1694,6 +1779,13 @@ async def _initialize(self) -> None:
...
                 await self._initialize_once()
+                try:
+                    await self.detach()
+                except Exception:
+                    logger.warning(
+                        "Failed to detach Playwright resources after cloud browser initialization",
+                        exc_info=True,
+                    )
                 return
             except Exception as error:
```
zizhengtai, line 2187: Could we clean up the cloud session and Playwright client when initialization is cancelled? `asyncio.CancelledError` bypasses this `except Exception`, so cancelling or timing out `env.start()` after session creation skips `_cleanup_failed_initialization_attempt()`: the cloud session remains allocated and the Playwright driver pipe stays open. Please handle cancellation separately, await cleanup, and re-raise without retrying. A regression test can cancel `_initialize_once()` after it stores the session/client and assert both session cleanup and `__aexit__()` occur.

zizhengtai (commented):
Two additional lifecycle issues from the second-pass review: refresh time-sensitive cloud CDP credentials on reconnect, and clean up cloud initialization on cancellation.

zizhengtai (approved):
Reviewed the latest branch changes. The lifecycle fixes are focused, the simplifications keep the implementation straightforward, and all checks pass.

PR #169, opened 2026-07-10
wait for extension to be autoloaded on Windows

volodymyr-narada:


packages/narada/src/narada/environment.py
```diff
@@ -1625,11 +1628,23 @@ async def _wait_for_browser_window_id_with_lazy_login(
                     if not config.interactive:
                         raise
 
-                    self._console.input(
-                        "\n[bold]>[/bold] [bold blue]The Narada Enterprise extension is not "
-                        "installed. Please follow the instructions in the browser window to "
-                        "install it first, then press Enter to continue.[/bold blue]\n",
-                    )
+                    # A user may configure the extension to be installed automatically through the
```
zizhengtai, line 1628: Add a TODO documenting the limitation of this approach: users who did install through registry and ran into this delay, will see a false install extension dialog on the page while the retry logic is happening. It's misleading as they can't manually download the extension by clicking the button. We should see if we can detect whether the extension is installed through the registry.

zizhengtai (approved):

packages/narada/src/narada/environment.py
```diff
@@ -1625,11 +1628,23 @@ async def _wait_for_browser_window_id_with_lazy_login(
                     if not config.interactive:
```
zizhengtai: This should be after the extension missing retry attempt below. In non-interactive mode, we should silently retry and only re-raise after all retry attempts fail.

zizhengtai (approved):

PR #171, opened 2026-07-14
check if extension is autoinstalled

volodymyr-narada:


packages/narada/src/narada/config.py
```diff
@@ -5,7 +5,7 @@
 
 def _default_executable_path() -> str:
     if sys.platform == "win32":
-        return "C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe"
+        return "C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe"
```
volodymyr-narada: TODO: remove

packages/narada/src/narada/environment.py
```diff
@@ -10,6 +10,12 @@
 import subprocess
 import sys
 import time
+
+try:
+    import winreg
```
volodymyr-narada: try-except because used only on Windows

zizhengtai (approved):

PR #172, opened 2026-07-14
retry chrome window if 1st launch fails when using extension autoload

volodymyr-narada:


packages/narada/src/narada/environment.py
```diff
@@ -126,6 +127,12 @@ class _BrowserInitializationResult(TypedDict, total=False):
     browserWindowId: str
 
 
+class _BrowserAutoloadRestartRequired(NaradaInitializationError):
```
zizhengtai: I think no reason to make it inherit from `NaradaInitializationError`, it might actually be more error-prone. Let's just make an independent exception.

packages/narada/src/narada/environment.py
```diff
@@ -1707,7 +1843,11 @@ async def _wait_for_browser_window_id_with_lazy_login(
                         wait_until="domcontentloaded",
                     )
 
-        except PlaywrightError:
+        except PlaywrightError as error:
```
zizhengtai, line 1846: Can we print this error always, for easier debugging?

zizhengtai (approved):

PR #174, opened 2026-07-15
Preserve operator action trace timing

soon-dubu:
## Summary

- Add optional Operator action trace fields for `credits`, `startTs`, `endTs`, `durationMs`, and nested `children`.
- Preserve backend-provided timing through both CPython and Pyodide SDK responses.
- Remove the earlier SDK-side workflow trace duration enrichment and its tests.

## Testing

- `uvx ruff check packages/narada-core/src/narada_core/models.py packages/narada-core/src/narada_core/tracing/model.py packages/narada/tests/test_cloud_browser.py packages/narada-pyodide/tests/test_cloud_browser.py`
- `uv run pytest packages/narada/tests/test_cloud_browser.py -k action_trace_timing`
- `uv run pytest packages/narada-pyodide/tests/test_cloud_browser.py -k action_trace_timing`
- `git diff --check`

## Codex sibling refs

- frontend: https://github.com/NaradaAI/frontend/pull/1944
- caddie: https://github.com/NaradaAI/caddie/pull/6775
- architecture-docs: https://github.com/NaradaAI/architecture-docs/pull/55

### DIFF
--- a/packages/narada-core/src/narada_core/tracing/model.py
+++ b/packages/narada-core/src/narada_core/tracing/model.py
@@ -44,7 +44,10 @@ class OperatorActionTraceItem(BaseModel):
             return self
         if self.end_ts < self.start_ts:
             raise ValueError("endTs must be greater than or equal to startTs")
-        if self.duration_ms is not None and self.duration_ms != self.end_ts - self.start_ts:
+        if (
+            self.duration_ms is not None
+            and self.duration_ms != self.end_ts - self.start_ts
+        ):
             raise ValueError("durationMs must equal endTs minus startTs")
         return self
 
--- a/packages/narada-core/src/narada_core/tracing/model.py
+++ b/packages/narada-core/src/narada_core/tracing/model.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, Self
 
 from pydantic import (
     BaseModel,
+    ConfigDict,
     Field,
     NonNegativeInt,
     TypeAdapter,
@@ -27,8 +28,25 @@ def _normalize_agent_type(agent_type: object) -> str:
 
 
 class OperatorActionTraceItem(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+
     url: str
     action: str
+    credits: float | None = None
+    start_ts: NonNegativeInt | None = Field(default=None, alias="startTs")
+    end_ts: NonNegativeInt | None = Field(default=None, alias="endTs")
+    duration_ms: NonNegativeInt | None = Field(default=None, alias="durationMs")
+    children: list[OperatorActionTraceItem] | None = None
+
+    @model_validator(mode="after")
+    def validate_timing(self) -> Self:
+        if self.start_ts is None or self.end_ts is None:
+            return self
+        if self.end_ts < self.start_ts:
+            raise ValueError("endTs must be greater than or equal to startTs")
+        if self.duration_ms is not None and self.duration_ms != self.end_ts - self.start_ts:
+            raise ValueError("durationMs must equal endTs minus startTs")
+        return self
 
 
 class GoToUrlTrace(BaseModel):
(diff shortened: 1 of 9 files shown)


packages/narada-core/src/narada_core/tracing/model.py
```diff
@@ -29,6 +29,11 @@ def _normalize_agent_type(agent_type: object) -> str:
 class OperatorActionTraceItem(BaseModel):
     url: str
     action: str
+    credits: float | None = None
+    startTs: NonNegativeInt | None = None
```
xTRam1: [IMPORTANT] These Pydantic objects are returned through the public Python `AgentResponse.action_trace` API, so using the wire casing here leaks a TypeScript/JSON representation into Python user code: callers get `item.startTs`, while `item.start_ts` does not exist. This boundary already translates `workflowTrace` to `workflow_trace`; the new fields should follow the same pattern before they become a released compatibility obligation.

Please keep camelCase in the wire-facing `TypedDict`, but expose snake_case attributes from this model with aliases:

```py
class OperatorActionTraceItem(BaseModel):
    model_config = ConfigDict(populate_by_name=True)

    url: str
    action: str
    credits: float | None = None
    start_ts: NonNegativeInt | None = Field(default=None, alias="startTs")
    end_ts: NonNegativeInt | None = Field(default=None, alias="endTs")
    duration_ms: NonNegativeInt | None = Field(default=None, alias="durationMs")
    children: list["OperatorActionTraceItem"] | None = None
```

Please test both sides of the abstraction: Python attribute access uses snake_case, while `model_dump(by_alias=True, exclude_none=True)` reproduces the backend's camelCase payload recursively.

packages/narada-core/src/narada_core/tracing/model.py
```diff
@@ -29,6 +29,11 @@ def _normalize_agent_type(agent_type: object) -> str:
...
+    endTs: NonNegativeInt | None = None
+    durationMs: NonNegativeInt | None = None
```
xTRam1: [IMPORTANT] These three independent nonnegative fields still accept an impossible interval such as `startTs=200, endTs=100, durationMs=999`. That contradicts the trace contract and is inconsistent with the ordering validators already used by `PythonSubAgentCallEvent` and `PythonExtensionActionEvent` in this module.

Please add a conditional model validator: when both endpoints are present, require `end_ts >= start_ts`; when all three values are present, either require `duration_ms == end_ts - start_ts` or explicitly define why duration has different semantics. For the current frontend/Caddie contract, equality is the intended invariant.

Include negative tests for reversed endpoints and a mismatched duration, plus a compatibility test showing that legacy rows with all timing fields absent still parse.

xTRam1 (changes requested):
Requesting changes before this public SDK surface is released.

The good parts are substantial: the CPython and Pyodide paths remain in parity, nested `children` are preserved recursively, old traces remain compatible because the fields are optional, and both focused runtime tests pass locally.

Three items remain:

1. Expose Pythonic snake_case attributes with camelCase wire aliases.
2. Validate ordering and duration consistency as one timing interval.
3. Rebase onto current `main` and bump releasable package metadata. Current main is already tagged/released as `narada-core 0.1.3`, `narada 0.2.6`, and `narada-pyodide 0.1.3`; merging this new public behavior under those same versions would make it impossible to publish cleanly. Bump all three to new unreleased versions, update the exact `narada-core` pins in both consumer packages, and regenerate `uv.lock`.

The version files are unchanged in this diff, so that release requirement is recorded here rather than on an unrelated changed line. All current SDK checks are green; this verdict is about the public API and release contract.

soon-dubu:
Addressed: the Python model exposes snake_case timing attributes with camelCase wire aliases, validates complete intervals, and the rebased branch bumps all three unreleased package versions and lockfile pins.

xTRam1 (approved):
Re-reviewed after the requested fixes. The public model now exposes Pythonic attributes with wire aliases, validates timing as one coherent interval, preserves recursive serialization and legacy untimed rows, and the full CI suite is green. Backward parsing was also verified against published narada-core 0.1.3. Ready to merge.

PR #181, opened 2026-07-25
Add OpenAI-shaped response trace models

soon-dubu:
## Summary

- define OpenAI-shaped trace and span models for workflows, agents, GUI/control flow, and agent actions
- make `AgentActionSpanData` message-only: persisted user-facing copy lives in `message`, while failures live in the enclosing span's `error`
- add nullable inclusive credits to workflow, agent, and action span data
- define positive, zero, and null settlement semantics and warn that inclusive tree values must not be summed
- keep untimed multipleActions children structural with null timestamps and null credits
- remove trace-specific UsageData while preserving AgentResponse.usage and AgentUsage
- retain workflow inputs, effective reasoning, starting_url, and independent group_id semantics
- keep legacy action_trace and workflow_trace unchanged

## Testing

- response trace model suite: 31 passed
- Ruff check and format passed

## Codex sibling refs

- caddie: https://github.com/NaradaAI/caddie/pull/6887
- frontend: https://github.com/NaradaAI/frontend/pull/2001
- architecture-docs: https://github.com/NaradaAI/architecture-docs/pull/74
- desktop-automation-app:
- api-docs:


packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -438,19 +438,27 @@ class FinallySpanData(ControlFlowSpanDataBase):
...
+class BaseAgentSpanData(SpanData):
     type: str
     name: str
-    handoffs: list[str] | None = None
-    tools: list[str] | None = None
     output_type: str | None = None
+    additional_tools: list[str] = Field(default_factory=list)
+    attachments: list[str] = Field(default_factory=list)
+    vector_stores: list[str] = Field(default_factory=list)
+    output_variables: list[dict[str, Any]] = Field(default_factory=list)
```
soon-dubu: This should be output_variables: dict[str, Any]

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    workflow_run_id: str | None = None
+    status: SpanStatus
+    request_id: str | None = None
+    input_summary: str | None = None
+    output_summary: str | None = None
+    output_names: list[str] | None = None
+    termination_mode: Literal["completed", "end_tree"] | None = None
+
+
+class GuiStepSpanDataBase(SpanData):
```
soon-dubu: Rename to BaseGuiStepSpanData

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+
+
+class TraceMetadata(BaseModel):
+    schema_version: Literal["1"] = "1"
+
+
+class Trace(BaseModel):
+    object: Literal["trace"] = "trace"
+    id: str
+    workflow_name: str
```
soon-dubu: No workflow name since it should be a part of a workflow span

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    group_id: str | None = None
+    metadata: TraceMetadata = Field(default_factory=TraceMetadata)
```
soon-dubu: Change to optional dictionary

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+        if (
+            self.started_at is not None
+            and self.ended_at is not None
+            and self.ended_at < self.started_at
+        ):
+            raise ValueError("ended_at must be greater than or equal to started_at")
+        return self
+
+
+type TraceRecord = Annotated[Trace | Span, Field(discriminator="object")]
```
soon-dubu: Confirm that OpenAI calls this union a TraceRecord

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+
+class SpanData(_OmitNoneModel):
+    type: str
+
+
+class WorkflowSpanData(SpanData):
+    type: Literal["workflow"] = "workflow"
+    name: str
+    workflow_id: str
+    workflow_run_id: str | None = None
```
soon-dubu: Remove this, request_id serves this purpose

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    data: dict[str, JsonValue] | None = None
+
+
+class SpanData(_OmitNoneModel):
+    type: str
+
+
+class WorkflowSpanData(SpanData):
+    type: Literal["workflow"] = "workflow"
+    name: str
```
soon-dubu: Change to workflow_name

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    "error",
+    "aborted",
+    "input_required",
+    "skipped",
+    "timeout",
+    "unknown",
+]
+
+
+class _OmitNoneModel(BaseModel):
```
soon-dubu: Think about this more, does OpenAI do this?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+
+
+class WorkflowSpanData(SpanData):
+    type: Literal["workflow"] = "workflow"
+    name: str
+    workflow_id: str
+    workflow_run_id: str | None = None
+    status: SpanStatus
+    request_id: str | None = None
+    input_summary: str | None = None
```
soon-dubu: Look into this

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+class SpanData(_OmitNoneModel):
+    type: str
+
+
+class WorkflowSpanData(SpanData):
+    type: Literal["workflow"] = "workflow"
+    name: str
+    workflow_id: str
+    workflow_run_id: str | None = None
+    status: SpanStatus
```
soon-dubu: should be whatever status is stored in the remote dispatch row

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    request_id: str | None = None
+    input_summary: str | None = None
+    output_summary: str | None = None
+    output_names: list[str] | None = None
+    termination_mode: Literal["completed", "end_tree"] | None = None
```
soon-dubu: get rid of this

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+class WorkflowSpanData(SpanData):
+    type: Literal["workflow"] = "workflow"
+    name: str
+    workflow_id: str
+    workflow_run_id: str | None = None
+    status: SpanStatus
+    request_id: str | None = None
+    input_summary: str | None = None
+    output_summary: str | None = None
+    output_names: list[str] | None = None
```
soon-dubu: what is this?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
+from __future__ import annotations
```
soon-dubu: add descriptions for each field

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    SerializerFunctionWrapHandler,
+    field_validator,
+    model_serializer,
+    model_validator,
+)
+
+type JsonValue = (
+    str | int | float | bool | None | list["JsonValue"] | dict[str, "JsonValue"]
+)
+type SpanStatus = Literal[
```
soon-dubu: we need differen statuses/no status for different kinds of spans

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    termination_mode: Literal["completed", "end_tree"] | None = None
+
+
+class GuiStepSpanDataBase(SpanData):
+    type: str
+    name: str | None = None
+    step_id: str
+    status: SpanStatus
+    description: str | None = None
+    page_url: str | None = None
```
soon-dubu: this is the url that we get before reaching this step?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+
+class AgentStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.agent"] = "gui_step.agent"
+    selected_agent_type: str
+
+
+class AgenticMouseActionStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.agentic_mouse_action"] = "gui_step.agentic_mouse_action"
+    strategy: Literal["direct", "operator_fallback"]
+    verification_status: bool | None = None
```
soon-dubu: look into this, what is it?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    type: Literal["gui_step.read_local_filesystem"] = "gui_step.read_local_filesystem"
+
+
+class WriteLocalFilesystemStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.write_local_filesystem"] = "gui_step.write_local_filesystem"
+
+
+class EndStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.end"] = "gui_step.end"
+    termination_mode: Literal["end", "end_tree"]
```
soon-dubu: get rid of this

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    result_status: Literal["success", "error"] | None = None
+    message: str | None = None
+
+
+class ForStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.for"] = "gui_step.for"
+    loop_type: Literal[
```
soon-dubu: can get rid of loop_type

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    type: Literal["gui_step.get_screenshot"] = "gui_step.get_screenshot"
+
+
+class GetSimplifiedHtmlStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.get_simplified_html"] = "gui_step.get_simplified_html"
+
+
+class GetUrlStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.get_url"] = "gui_step.get_url"
+    observed_url: str | None = None
```
soon-dubu: look into this, what is it?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    type: Literal["gui_step.write_google_sheet"] = "gui_step.write_google_sheet"
+
+
+class RunCustomAgentStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.run_custom_agent"] = "gui_step.run_custom_agent"
+    workflow_id: str
+    workflow_name: str
+
+
+class RunCustomAgentsInParallelStepData(GuiStepSpanDataBase):
```
soon-dubu: Get rid of this

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    type: Literal["gui_step.run_custom_agents_in_parallel"] = (
+        "gui_step.run_custom_agents_in_parallel"
+    )
+
+
+class RunCustomAgentForEachStepData(GuiStepSpanDataBase):
```
soon-dubu: Get rid of this

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+        "gui_step.run_custom_agent_for_each"
+    )
+    workflow_id: str
+    workflow_name: str
+    total_items: NonNegativeInt
+
+
+class OutputStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.output"] = "gui_step.output"
+    output_names: list[str] | None = None
```
soon-dubu: some kind of dict

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    | RunCustomAgentStepData
+    | RunCustomAgentsInParallelStepData
+    | RunCustomAgentForEachStepData
+    | OutputStepData
+    | CriticAgentStepData,
+    Field(discriminator="type"),
+]
+
+
+class ControlFlowSpanDataBase(SpanData):
```
soon-dubu: Rename to BaseControlFlowSpanData

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    type: Literal["control_flow.finally"] = "control_flow.finally"
+
+
+type ControlFlowSpanData = Annotated[
+    IterationSpanData | TrySpanData | CatchSpanData | FinallySpanData,
+    Field(discriminator="type"),
+]
+
+
+class UsageData(_OmitNoneModel):
```
soon-dubu: keep actions, also explain to me what usagedata is

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    credits: NonNegativeFloat
+    input_tokens: NonNegativeInt | None = None
```
soon-dubu: get ridn of tokens

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    output_variables: list[dict[str, Any]] = Field(default_factory=list)
+    reasoning_effort: Literal[
+        "agent_default",
+        "none",
+        "low",
+        "medium",
+        "high",
+    ] = "agent_default"
+    status: SpanStatus
+    agent_id: str | None = None
```
soon-dubu: what is agent_id?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+        self,
+        handler: SerializerFunctionWrapHandler,
+    ) -> Any:
+        serialized = handler(self)
+        if not isinstance(serialized, dict):
+            return serialized
+        return {key: value for key, value in serialized.items() if value is not None}
+
+
+class TraceMetadata(BaseModel):
```
soon-dubu: Get rid of this

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+class AgentStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.agent"] = "gui_step.agent"
+    selected_agent_type: str
+
+
+class AgenticMouseActionStepData(GuiStepSpanDataBase):
+    type: Literal["gui_step.agentic_mouse_action"] = "gui_step.agentic_mouse_action"
+    strategy: Literal["direct", "operator_fallback"]
+    verification_status: bool | None = None
+    verification_status_variable_name: str | None = None
```
soon-dubu: what is this as well?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    input_tokens: NonNegativeInt | None = None
+    output_tokens: NonNegativeInt | None = None
+    total_tokens: NonNegativeInt | None = None
+
+
+class BaseAgentSpanData(SpanData):
+    type: str
+    name: str
+    output_type: str | None = None
+    additional_tools: list[str] = Field(default_factory=list)
```
soon-dubu: actionally don't need additional_tools, attachments, vector_stores

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+    attachments: list[str] = Field(default_factory=list)
+    vector_stores: list[str] = Field(default_factory=list)
+    output_variables: list[dict[str, Any]] = Field(default_factory=list)
+    reasoning_effort: Literal[
```
soon-dubu: don't need reasoning effort either

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+        "none",
+        "low",
+        "medium",
+        "high",
+    ] = "agent_default"
+    status: SpanStatus
+    agent_id: str | None = None
+    request_id: str | None = None
+    page_url: str | None = None
+    input_summary: str | None = None
```
soon-dubu: what are input/output summaries? same as workflow span

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,654 @@
...
+        "agent_default",
+        "none",
+        "low",
+        "medium",
+        "high",
+    ] = "agent_default"
+    status: SpanStatus
+    agent_id: str | None = None
+    request_id: str | None = None
+    page_url: str | None = None
```
soon-dubu: what are page_urls? same for everything that has this field

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `workflow_name` from the trace record. The executed workflow name now lives only on `WorkflowSpanData`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Replaced the generic status with `WorkflowSpanStatus`, matching remote-dispatch row values: `pending`, `input-required`, `success`, `error`, and `expired`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated `output_variables` to `dict[str, Any]` with runtime values rather than a list of variable definitions.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Confirmed: the OpenAI Agents SDK does not define or name a `Trace | Span` union as `TraceRecord`. I renamed our flattened-response union to the Narada-local `TraceItem`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: `verification_status` is the runtime result of post-action verification. It is absent if verification did not run. I kept that result, documented it, and removed the configured destination variable name.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `input_summary` and `output_summary`. The workflow payload now carries actual runtime `output_variables` when available instead of speculative summaries.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `termination_mode` from the workflow span.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Changed `metadata` to `dict[str, Any] | None`, defaulting to `None`. The field description reserves `schema_version` for Narada-owned schema metadata.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Renamed this base class to `BaseGuiStepSpanData`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: This represented authored output-variable names, so I removed it. The replacement is `output_variables: dict[str, Any] | None`, containing actual runtime output values when available.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `workflow_run_id`. `request_id` is now documented as the remote-dispatch lookup identifier for the workflow run.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Split status by source: workflow spans use remote-dispatch statuses, GUI steps use GUI execution statuses, and agent spans use agent-run statuses. Synthetic control-flow spans and agent-action spans no longer invent a status.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: This is the URL returned by the get-URL step at runtime. I renamed it from `observed_url` to `url` and added that definition to the field description.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `additional_tools`, `attachments`, and `vector_stores` from agent span data.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Yes. The current GUI evaluator captures this URL immediately before executing the step. I renamed it to `page_url_before` and documented that timing explicitly.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: I removed the authored per-workflow fields, but kept the `gui_step.run_custom_agents_in_parallel` subtype because the agreed hierarchy preserves every executed GUI step and nests child workflow spans below it. Did you intend to remove the entire GUI-step span for this step type?

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Added Pydantic `Field(description=...)` text to every public model field, plus a test that fails if any exported model field lacks a description.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `reasoning_effort` from agent span data.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: OpenAI does not use an omit-None base model for these exported records. I removed `_OmitNoneModel` and the agent-specific serializer, so normal Pydantic serialization preserves null optional fields.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `loop_type`; `total_iterations` remains because it is learned at runtime.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Renamed this field to `workflow_name`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `termination_mode` from the end-step payload.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Changed this to `output_variables: dict[str, Any]`, containing the runtime names and values emitted by the output step.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed all token fields. `UsageData` now contains only aggregate `actions` and `credits`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: `UsageData` is the aggregate billable usage recorded for an agent run. It now contains only `actions` and `credits`; `actions` is retained because the billed aggregate can differ from the action spans returned to the caller.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: I removed the authored workflow identity fields and kept only runtime `total_items`, but retained the `gui_step.run_custom_agent_for_each` subtype so the launching GUI step remains above its iteration and workflow spans. Did you intend to remove that entire GUI-step span?

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: This was the configured variable receiving the verification result, not execution-level information. I removed it and retained only the runtime `verification_status` result.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: They were undefined display summaries rather than concrete execution output. I removed both and retained `output_variables` for actual structured runtime output.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed the dedicated `TraceMetadata` model. `Trace.metadata` is now an optional dictionary.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: This was intended as a stable agent-definition identifier, but it is not consistently available as an execution result. I removed it.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed the aggregate `page_url` from agent spans. GUI steps now expose `page_url_before`, while individual agent actions retain `url` for the page associated with that action; both fields now have explicit descriptions.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Renamed this base class to `BaseControlFlowSpanData`.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -8,347 +8,553 @@
...
-    object: Literal["trace"] = "trace"
-    id: str
-    workflow_name: str
-    group_id: str | None = None
-    metadata: TraceMetadata = Field(default_factory=TraceMetadata)
+    object: Literal["trace"] = Field(
+        default="trace",
+        description="Discriminator identifying this record as a trace.",
+    )
+    id: str = Field(description="Unique identifier for the trace.")
```
soon-dubu: Should be named trace_id

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -539,14 +774,31 @@ class AgentActionSpanData(SpanData):
...
-    parent_id: str | None = None
-    started_at: datetime | None = None
-    ended_at: datetime | None = None
-    span_data: SpanDataUnion
-    error: SpanError | None = None
+    object: Literal["trace.span"] = Field(
+        default="trace.span",
+        description="Discriminator identifying this record as a span.",
+    )
+    id: str = Field(description="Unique identifier for the span.")
```
soon-dubu: this should be span_id

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -8,347 +8,553 @@
...
+    id: str = Field(description="Unique identifier for the trace.")
+    group_id: str | None = Field(
+        default=None,
+        description="Optional identifier used to correlate related traces.",
+    )
+    metadata: dict[str, Any] | None = Field(
+        default=None,
+        description=(
+            "Optional trace-level metadata. Narada-owned metadata uses "
+            "schema_version to identify the response trace schema."
```
soon-dubu: Should not be mentioning schema_version

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Look into the actual openai sdk code to understand. This should have the same name as openai's sdk

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: page_url is fine

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: RunCustomAgentsInParallelStep is no longer supported so you can get rid of this

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: We should standardize the naming to just page_url

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -8,347 +8,553 @@
...
+    type: Literal["gui_step.run_custom_agents_in_parallel"] = Field(
+        default="gui_step.run_custom_agents_in_parallel",
+        description="Identifies a GUI parallel custom-agent step.",
     )
 
 
-class RunCustomAgentForEachStepData(GuiStepSpanDataBase):
-    type: Literal["gui_step.run_custom_agent_for_each"] = (
-        "gui_step.run_custom_agent_for_each"
+class RunCustomAgentForEachStepData(BaseGuiStepSpanData):
```
soon-dubu: Get rid of RunCustomAgentForEachStepData totally

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -8,347 +8,553 @@
...
 
 
-class RunCustomAgentsInParallelStepData(GuiStepSpanDataBase):
-    type: Literal["gui_step.run_custom_agents_in_parallel"] = (
-        "gui_step.run_custom_agents_in_parallel"
+class RunCustomAgentsInParallelStepData(BaseGuiStepSpanData):
+    type: Literal["gui_step.run_custom_agents_in_parallel"] = Field(
+        default="gui_step.run_custom_agents_in_parallel",
+        description="Identifies a GUI parallel custom-agent step.",
     )
```
soon-dubu: Get rid of RunCustomAgentsInParallelStepData totally

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: I checked the current OpenAI source. It does not name this union at all: processor/exporter signatures use `list[Trace | Span[Any]]` directly. Per the latest direction, I removed both `TraceItem` and `TraceRecord`, made `Span` generic, and use the direct union in the test that defines the flattened response boundary.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: The later review clarified that this step is no longer supported. I removed `RunCustomAgentForEachStepData` completely, including its union and public exports.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Renamed the Python field to `span_id`. Its serialization alias is `id`, matching OpenAI's `Span.span_id` property and `SpanImpl.export()` output.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: I revisited this against the current OpenAI source. OpenAI's Python `Trace` exposes `name`, and `TraceImpl.export()` emits it as `workflow_name`. To satisfy the latest compatibility direction, Narada now does the same through a serialization alias: Python callers use `trace.name`; returned JSON contains `workflow_name`. The workflow span still has its own `workflow_name` because it identifies the executed Narada workflow span.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed the schema-version wording. `metadata` is now described only as optional user-provided trace metadata, matching OpenAI.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Renamed the Python field to `trace_id`. To match OpenAI on both surfaces, its serialization alias is `id`, so `trace.trace_id` is the Python API and `model_dump()` emits the OpenAI wire key `id`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Reverted this to `page_url`. The description still clarifies that the GUI evaluator captures it immediately before the step starts.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `RunCustomAgentsInParallelStepData` completely, including its union and public exports.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Source audit confirmation: OpenAI does not use an omit-None model. Its `TraceImpl.export()`, `SpanImpl.export()`, and standard span-data `export()` methods construct dictionaries explicitly and retain nullable keys. The Narada models therefore continue to use normal Pydantic serialization with nulls preserved.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `RunCustomAgentsInParallelStepData` entirely from the module, union, exports, and taxonomy tests.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Standardized the GUI-step field back to `page_url`. Agent action spans already use the same name.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed `RunCustomAgentForEachStepData` entirely from the module, union, exports, and taxonomy tests.

soon-dubu:
Re-audited this model against the current OpenAI Agents SDK source rather than relying on the prose guide:

- [`Trace` / `TraceImpl`](https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/traces.py): Python uses `trace_id` and `name`; export uses `id` and `workflow_name`.
- [`Span` / `SpanImpl`](https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/spans.py): Python uses `span_id`, `Span` is generic, and export uses `id` while retaining the standard nullable envelope fields.
- [`SpanData` / `AgentSpanData`](https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/span_data.py): `AgentSpanData` is a concrete class and its discriminator is `type="agent"`.
- [`TracingExporter`](https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/processor_interface.py): collections are typed directly as `list[Trace | Span[Any]]`; OpenAI defines neither `TraceItem` nor `TraceRecord`.

Commit 9b7292e now follows those conventions. Pydantic serialization aliases preserve OpenAI’s exported JSON keys while the Python attributes match OpenAI’s object API. Narada-only span data retains only the execution-level fields discussed in review.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+
+type WorkflowSpanStatus = Literal[
+    "pending",
+    "input-required",
+    "success",
+    "error",
+    "expired",
+]
+type GuiStepSpanStatus = Literal["success", "error", "aborted", "end_tree"]
+type AgentSpanStatus = Literal[
```
soon-dubu, line 76: Check that these types actually exist in remote dispatch table

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+    "success",
+    "error",
+    "aborted",
+    "input-required",
+    "timeout",
+]
+type AgentType = Literal[
+    "operator", "core", "productivity", "custom", "critic", "other"
```
soon-dubu: custom doesn't exist neither does other, run custom agent runs a workflow

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+
+class Trace(BaseModel):
+    model_config = ConfigDict(validate_by_name=True, serialize_by_alias=True)
+
+    object: Literal["trace"] = Field(
+        default="trace",
+        description="Discriminator identifying this record as a trace.",
+    )
+    trace_id: str = Field(
+        alias="id",
```
soon-dubu: Why do we have aliases? Do we need them?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+        description="Unique identifier for the trace.",
+    )
+    name: str = Field(
+        alias="workflow_name",
```
soon-dubu: Same here

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+
+class Span(BaseModel, Generic[TSpanData]):
+    model_config = ConfigDict(validate_by_name=True, serialize_by_alias=True)
+
+    object: Literal["trace.span"] = Field(
+        default="trace.span",
+        description="Discriminator identifying this record as a span.",
+    )
+    span_id: str = Field(
+        alias="id",
```
soon-dubu: PR wide comment: why do we need aliases everywhere

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+    WorkflowSpanData
+    | GuiStepSpanData
+    | ControlFlowSpanData
+    | AgentSpanData
+    | AgentActionSpanData,
+    Field(discriminator="type"),
+]
+
+
+class Span(BaseModel, Generic[TSpanData]):
```
soon-dubu: Why do we use a generic here? There should be an actual reason why openai uses this as opposed to a discriminated union. Look into this and find out why.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+        default="gui_step.get_url",
+        description="Identifies a GUI get-URL step.",
+    )
+    url: str | None = Field(
+        default=None,
+        description="URL returned by the step during execution.",
+    )
+
+
+class NavigateStepData(BaseGuiStepSpanData):
```
soon-dubu: Change to GoToUrlStepData. Each navigate step should be renamed to gotourl. PR wide comment: the names of the step data and their types should match what the steps are actually named.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+        description="HTTP response status code observed at runtime.",
+    )
+
+
+class IfStepData(BaseGuiStepSpanData):
+    type: Literal["gui_step.if"] = Field(
+        default="gui_step.if",
+        description="Identifies a GUI conditional step.",
+    )
+    selected_branch_role: Literal["then", "else_if", "else"] | None = Field(
```
soon-dubu: Instead of using selected_branch_role or selected_branch_index, let's use the selected condition, which should be a string. Any variables should not already be evaluated in this string.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,846 @@
...
+        description="Identifies a GUI throw step.",
+    )
+
+
+class TryCatchStepData(BaseGuiStepSpanData):
+    type: Literal["gui_step.try_catch"] = Field(
+        default="gui_step.try_catch",
+        description="Identifies a GUI try/catch step.",
+    )
+    caught_error: bool = Field(
```
soon-dubu: instead of these fields (caught_error, executed_catch, executed_finally), let's put in the first condition string that was caught, similar to the if statement from before. This should be none if nothing is caught.

packages/narada-pyodide/src/narada/tracing.py
```diff
@@ -0,0 +1,2 @@
+from narada_core.tracing.response_trace import *  # noqa: F403
+from narada_core.tracing.response_trace import __all__ as __all__
```
soon-dubu: why do we do this as well as import *

packages/narada/src/narada/tracing.py
```diff
@@ -0,0 +1,2 @@
+from narada_core.tracing.response_trace import *  # noqa: F403
+from narada_core.tracing.response_trace import __all__ as __all__
```
soon-dubu: ditto

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I’ll cross-check every agent-type literal against the actual remote-dispatch routing table and remove or rename anything that is not a real runtime type.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Agreed. I’ll remove `custom` and `other`; running a custom agent will be represented as a workflow run rather than as an agent type.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I’ll audit this alias against the OpenAI Python-object and exported-JSON shapes and keep it only if the two public names genuinely differ.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I’ll apply the same alias audit here and remove the alias if it is not required for OpenAI compatibility.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged as a PR-wide change. I’ll review every alias, document the small set that is required by OpenAI’s object-versus-export naming, and remove the rest.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I’ll inspect how OpenAI uses the generic in its runtime API and typing, then either document the concrete reason it applies here or replace it with the discriminated-union type if it does not.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Agreed. I’ll rename this to `GoToUrlStepData` with the matching `goToUrl` discriminator and audit every GUI-step data class/discriminator against the canonical Agent Studio step names.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Agreed. I’ll replace the branch role/index fields with an optional selected-condition string and preserve the authored condition template rather than substituting runtime variable values.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Agreed. I’ll replace `caught_error`, `executed_catch`, and `executed_finally` with the first matched catch-condition string, using `None` when no catch condition matched.

packages/narada-pyodide/src/narada/tracing.py
soon-dubu: Acknowledged. The explicit import currently exposes the source module’s `__all__` in addition to performing the star re-export. I’ll verify whether that is a necessary public contract and simplify the wrapper if it is not.

packages/narada/src/narada/tracing.py
soon-dubu: Acknowledged. I’ll make the desktop SDK wrapper follow the same simplified re-export approach as the Pyodide package.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Yes, there is a gui step called Run Custom Agent. This is a gui step that should have a workflow span underneath it if successful.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Check that the other statuses also work too. As in these are statuses that actually exist somewhere.

soon-dubu: Updated in d86e7b0. The literals now match concrete runtime contracts: workflow statuses match `RemoteDispatchCurrentStatus`, GUI-step statuses match `NodeStatus` including `endTree`, and agent statuses match `PythonSubAgentCallEvent` / `AgentResponse`. This removes the unsupported agent `aborted` value and adds exact-set tests for all three status types.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in d86e7b0. `custom`, `other`, and the non-runtime `critic` agent type are removed; `AgentType` now contains the exact Agent Studio dispatch values. `gui_step.runCustomAgent` remains a GUI-step subtype, and its contract now states that a successful step parents the invoked `WorkflowSpanData` span. A hierarchy test covers that relationship through `parent_id`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed in d86e7b0. The OpenAI Python object exposes `trace_id`; it only changes that key to `id` inside its separate `export()` implementation. Since this stage returns the object model and does not implement the exporter, the alias was unnecessary.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed in d86e7b0. The model now uses OpenAI Python-object field names directly (`name` here); `workflow_name` belongs to the separate OpenAI export payload.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Addressed PR-wide in d86e7b0. All response-trace aliases and alias-specific model configuration are removed. `Trace` now serializes its object fields as `trace_id` and `name`, and `Span` as `span_id`, matching the OpenAI Python API rather than conflating it with OpenAI's separate exporter payload.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Kept and documented in d86e7b0 after checking the OpenAI source. OpenAI makes `Span` generic so a caller holding `Span[AgentSpanData]` retains the concrete type of `.span_data`; this matters when constructing or consuming typed spans. A discriminated union solves the separate problem of decoding serialized data whose subtype is unknown, so `SpanDataUnion` remains available for that path.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in d86e7b0. This is now `GoToUrlStepData` with `type="gui_step.goToUrl"`. I also audited every modeled GUI-step subtype against the canonical Agent Studio registry: discriminator suffixes now retain the exact camelCase step names, and the mismatched Python classes are now `PythonStepData` and `NaradaCodeProjectExecutableStepData`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in d86e7b0. `IfStepData` now has only `selected_condition: str | None`, described as the authored condition with variable references left unevaluated. It is null for an else branch or when no branch ran.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in d86e7b0. The three boolean fields are replaced by `caught_condition: str | None`, containing the authored condition for the first matching catch branch and remaining null when no error was caught.

packages/narada-pyodide/src/narada/tracing.py
soon-dubu: Simplified in d86e7b0. The star import already honors `response_trace.__all__`, so importing that same `__all__` into the thin Pyodide wrapper was redundant and is removed.

packages/narada/src/narada/tracing.py
soon-dubu: Simplified in d86e7b0. The desktop wrapper now uses only the star re-export, matching the Pyodide wrapper.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -129,8 +130,8 @@ class AgentStepData(BaseGuiStepSpanData):
 
 
 class AgenticMouseActionStepData(BaseGuiStepSpanData):
-    type: Literal["gui_step.agentic_mouse_action"] = Field(
-        default="gui_step.agentic_mouse_action",
+    type: Literal["gui_step.agenticMouseAction"] = Field(
```
soon-dubu: PR wide comment: types should remain snake case

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -20,32 +19,34 @@
     "error",
     "expired",
 ]
-type GuiStepSpanStatus = Literal["success", "error", "aborted", "end_tree"]
+type GuiStepSpanStatus = Literal["success", "error", "aborted", "endTree"]
```
soon-dubu: snake_case was fine for end_tree

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in 61679a4. All `type` discriminator values are snake_case again, including corrected names such as `gui_step.go_to_url`, `gui_step.narada_code_project_executable`, and `gui_step.run_custom_agent`. The Python class names still mirror the canonical GUI step names.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in 61679a4. `GuiStepSpanStatus` uses `end_tree` again, consistently with the snake_case public trace schema.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Follow-up scope correction in 61679a4: Python Agent Studio workflow events are no longer a supported source for this model, so `timeout` has been removed. `AgentSpanStatus` now exactly matches the local SDK `AgentResponse` contract: `success`, `error`, and `input-required`.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,852 @@
...
+        description="Identifier of the step in the workflow definition."
+    )
+    status: GuiStepSpanStatus = Field(
+        description="Terminal execution status reported for the GUI step."
+    )
+    description: str | None = Field(
+        default=None,
+        description="User-facing description produced while the step executed.",
+    )
+    page_url: str | None = Field(
```
soon-dubu: Rename to starting_url

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,852 @@
...
+            "the identifier used to look up the run."
+        ),
+    )
+    output_variables: dict[str, Any] | None = Field(
+        default=None,
+        description="Runtime output variables produced by the workflow, when available.",
+    )
+
+
+class BaseGuiStepSpanData(SpanData):
```
soon-dubu: Can we look into adding a step number field? Like in the gui agent studio, we can see what step number a gui step might be at.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in 8087c95: GUI step spans now expose `starting_url`, documented as the browser URL captured immediately before the step starts. I also removed the unsupported `final_url` field from `GoToUrlStepData` so the model does not invent a navigation output.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Added in dbaa498. `BaseGuiStepSpanData.step_number` is an optional positive integer representing the same one-based, depth-first display position that Agent Studio computes across top-level and nested steps. Keeping it optional preserves compatibility with existing trace producers that only have `step_id`.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -534,16 +1561,42 @@ class RunCustomAgentStepData(BaseGuiStepSpanData):
             "the workflow span for the invoked workflow."
         ),
     )
+    input: RunCustomAgentStepInput | None = Field(
+        default=None,
+        description="Effective custom-workflow input used by the step.",
+    )
+
+
+class RunCustomAgentsInParallelStepData(BaseGuiStepSpanData):
```
soon-dubu: We are not supporting RunCustomAgentsInParallel. Remove this.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -534,16 +1561,42 @@ class RunCustomAgentStepData(BaseGuiStepSpanData):
...
+        default="gui_step.run_custom_agents_in_parallel",
+        description="Identifies a GUI parallel custom-agent step.",
+    )
+    input: RunCustomAgentsInParallelStepInput | None = Field(
+        default=None,
+        description="Effective parallel custom-agent inputs used by the step.",
+    )
+
+
+class RunCustomAgentForEachStepData(BaseGuiStepSpanData):
```
soon-dubu: We are not supporting RunCustomAgentForEach. Remove this.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -100,6 +100,863 @@ class WorkflowSpanData(SpanData):
...
+        description="Workflow file variable written to the local filesystem."
+    )
+
+
+class EndStepInput(BaseModel):
+    terminate_tree: bool = Field(
+        default=False,
+        description="Whether the step terminates the complete workflow tree.",
+    )
+    result_status: Literal["success", "error"] | None = Field(
```
soon-dubu: When I look at the end step, the only thing I can see is terminate_tree. Neither result_status or message. Look into if these actually exist somewhere. If not, remove them.

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -100,6 +100,863 @@ class WorkflowSpanData(SpanData):
...
+        default=None,
+        description="Workflow result selected when the complete tree is terminated.",
+    )
+    message: str | None = Field(
```
soon-dubu: same as above

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I will remove the unsupported parallel custom-agent span and input types from the public taxonomy and tests.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I will remove the unsupported custom-agent-for-each span and input types from the public taxonomy and tests.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged. I am checking the End step definition, editor, compiler, and evaluator to distinguish stored legacy fields from controls users can currently configure. I will keep only fields present in the supported GUI surface.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Acknowledged together with the preceding End-step comment; I will remove `message` as well if it is not exposed by the supported GUI editor.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed the unsupported RunCustomAgentsInParallel input/data models, discriminator case, public exports, and taxonomy expectation in 332a556.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Removed the unsupported RunCustomAgentForEach input/data models, discriminator case, public exports, and taxonomy expectation in 332a556.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Confirmed these are supported conditional End-step inputs. Enabling terminateTree reveals “Return as error” (resultStatus) and “Termination message” (message) in EndStepForm, and EndStepConfig.evaluate consumes both values. I kept them in EndStepInput and added a regression test in 332a556.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Confirmed alongside the comment above: message is an optional conditional field shown when terminateTree is enabled and is interpolated by the End-step evaluator, so it remains in EndStepInput. Covered by the regression test in 332a556.

packages/narada-core/src/narada_core/tracing/span_data.py
```diff
@@ -0,0 +1,805 @@
...
+)
+
+type WorkflowSpanStatus = Literal[
+    "pending",
+    "input-required",
+    "success",
+    "error",
+    "expired",
+]
+type GuiStepSpanStatus = Literal["success", "error", "aborted", "end_tree"]
```
soon-dubu: Why is end_tree a potential status here?

packages/narada-core/src/narada_core/tracing/response_trace.py
```diff
@@ -0,0 +1,11 @@
+"""Compatibility exports for the response trace model."""
+
+from narada_core.tracing.span_data import *  # noqa: F403
```
soon-dubu: DMove all these imports into tracing.  Just have \_all\__ for __init__.py. Don't have \__all\__ for the individual files.

packages/narada-core/src/narada_core/tracing/span_data.py
```diff
@@ -0,0 +1,823 @@
...
+    "error",
+    "input-required",
+]
+
+
+class SpanData(BaseModel):
+    type: str = Field(description="Discriminator for the span-specific payload.")
+
+
+TGuiStepSpanInput = TypeVar("TGuiStepSpanInput", bound=BaseModel)
```
soon-dubu: Bound this by a discrimanted union of possible spaninputs defined in step_span_inputs.py

packages/narada-core/src/narada_core/tracing/span_data.py
```diff
@@ -0,0 +1,823 @@
...
+        description="Identifies a GUI end step.",
+    )
+
+
+class ForStepData(BaseGuiStepSpanData[ForStepSpanInput]):
+    type: Literal["gui_step.for"] = Field(
+        default="gui_step.for",
+        description="Identifies a GUI for-loop step.",
+    )
+    total_iterations: NonNegativeInt = Field(
```
soon-dubu: Get rid of total_iterations

packages/narada-core/src/narada_core/tracing/span_data.py
```diff
@@ -0,0 +1,823 @@
...
+        description="Identifies a GUI wait-for-element step.",
+    )
+
+
+class WhileStepData(BaseGuiStepSpanData[WhileStepSpanInput]):
+    type: Literal["gui_step.while"] = Field(
+        default="gui_step.while",
+        description="Identifies a GUI while-loop step.",
+    )
+    total_iterations: NonNegativeInt = Field(
```
soon-dubu: Get rid of total_iterations here as well

packages/narada-core/src/narada_core/tracing/span_data.py
soon-dubu: Already addressed in f3cb103: `GuiStepSpanStatus` now contains only `success`, `error`, and `aborted`, and the validation test rejects `end_tree`.

packages/narada-core/src/narada_core/tracing/response_trace.py
soon-dubu: Updated in 4522f20. The explicit public imports and sole `__all__` now live in `narada_core.tracing.__init__`; the individual model modules no longer define `__all__`, and the redundant `response_trace.py` facade is removed.

packages/narada-core/src/narada_core/tracing/span_data.py
soon-dubu: Updated in 4522f20. `step_span_inputs.py` now defines the `type`-discriminated `GuiStepSpanInput` union across all supported step-input snapshots, and `TGuiStepSpanInput` is bounded to that union.

packages/narada-core/src/narada_core/tracing/span_data.py
soon-dubu: Removed `total_iterations` from `ForStepData` in 4522f20 and added a contract assertion that the field is absent.

packages/narada-core/src/narada_core/tracing/span_data.py
soon-dubu: Removed `total_iterations` from `WhileStepData` in 4522f20 and covered its absence alongside the for-loop model.

packages/narada-core/src/narada_core/tracing/span_data.py
```diff
@@ -0,0 +1,729 @@
...
+
+class AgentSpanData(SpanData):
+    type: Literal["agent"] = Field(
+        default="agent",
+        description="Identifies an agent span, matching OpenAI's agent span type.",
+    )
+    name: str = Field(description="Display name of the agent that executed.")
+    agent_type: AgentType = Field(
+        description="Narada agent type selected for this execution."
+    )
```
soon-dubu, line 677: Should also have a "message" or "prompt" field, whatever we currently call the prompt that goes as input into an agent

soon-dubu: Acknowledged. The SDK entrypoint and remote-dispatch path consistently call this input `prompt`, so I will add a required `prompt: str` field to `AgentSpanData` and cover it in the serialization contract tests.

packages/narada-core/src/narada_core/tracing/span_data.py
soon-dubu: Acknowledged follow-up: making the input union independently discriminated duplicates the enclosing `span_data.type`. I am retaining `GuiStepSpanInput` as the finite `TGuiStepSpanInput` bound, but changing it to a plain union. The outer `GuiStepSpanData` discriminator selects the concrete step model, whose generic specialization then validates the corresponding nested input without an `input.type` field.

packages/narada-core/src/narada_core/tracing/span_data.py
soon-dubu: Added in 37321ca. `AgentSpanData` now requires `prompt: str`, matching `Agent.run(prompt=...)` and the remote-dispatch terminology. The serialization and status/agent-type contract tests now supply and verify the prompt.

packages/narada-core/src/narada_core/tracing/traces.py
```diff
@@ -18,7 +18,10 @@ class Trace(BaseModel):
     )
     group_id: str | None = Field(
         default=None,
-        description="Optional identifier used to correlate related traces.",
+        description=(
+            "Optional caller-supplied identifier used to correlate related traces. "
+            "This is distinct from per-run request identifiers."
```
soon-dubu: Can remove the per-run request identifier portion here

soon-dubu: Removed in c51ac36. `group_id` now only describes caller-supplied trace correlation.

packages/narada-core/tests/test_response_trace_models.py
```diff
@@ -300,20 +300,19 @@ def test_taxonomy_discriminators_are_complete() -> None:
...
     )
 
     assert action.model_dump(mode="json") == {
         "type": "agent_action",
-        "name": "click",
         "message": "Clicked Submit",
         "starting_url": "https://example.test/form",
         "credits": 4.0,
     }
     assert "url" not in AgentActionSpanData.model_fields
```
soon-dubu: assert "url" not in AgentActionSpanData.model_fields
assert "name" not in AgentActionSpanData.model_fields

Don't need these tests

soon-dubu: Removed both redundant negative field assertions in d18cc96. The positive serialization tests continue to cover the public action shape.

PR #185, opened 2026-08-15
Expose agent model tiers in the Python SDK

xTRam1:
## Summary

- add the public `AgentModelTier` enum with Default and Mini product tiers
- expose the tier on `Agent` in both CPython and Pyodide
- send `modelTier=mini` through remote dispatch while omitting Default for backward compatibility
- scope `model_tier` to direct built-in `AgentKind` invocations
- reject Mini for named Agent Studio workflows because their persisted Agent steps own model tiers; accepting it would otherwise be a silent no-op
- update package versions, dependency pins, lockfile, tests, and SDK usage documentation

## Compatibility and rollout

Default requests preserve the existing wire shape by omitting `modelTier`. Mini requires the linked Caddie remote-dispatch support. Publish `narada-core==0.1.7`, `narada-pyodide==0.1.8`, and `narada==0.2.10` before merging the standalone frontend browser pin: https://github.com/NaradaAI/frontend/pull/2066

## Testing

- CPython and Pyodide focused Agent tests: 45 passed
- Ruff lint/format and `git diff --check` passed
- isolated package builds passed
- independent final frontend/SDK boundary review: PASS

## Codex sibling refs

- caddie: https://github.com/NaradaAI/caddie/pull/6983
- frontend: https://github.com/NaradaAI/frontend/pull/2065
- desktop-automation-app:
- api-docs:

## Related contract docs

- https://github.com/NaradaAI/architecture-docs/pull/82

### DIFF
--- a/packages/narada-core/src/narada_core/models.py
+++ b/packages/narada-core/src/narada_core/models.py
@@ -23,6 +23,13 @@ class AgentKind(Enum):
                 return "/coreAgent "
 
 
+class AgentModelTier(StrEnum):
+    """Selects the product model tier for a Narada agent invocation."""
+
+    DEFAULT = "default"
+    MINI = "mini"
+
+
 class ReasoningEffort(StrEnum):
     """Controls how much reasoning a built-in Narada agent uses before responding."""
 
--- a/packages/narada-pyodide/pyproject.toml
+++ b/packages/narada-pyodide/pyproject.toml
@@ -1,14 +1,14 @@
 
 [project]
 name = "narada-pyodide"
-version = "0.1.7"
+version = "0.1.8"
 description = "Pyodide-compatible Python client SDK for Narada"
 license = "Apache-2.0"
 readme = "README.md"
 authors = [{ name = "Narada", email = "support@narada.ai" }]
 requires-python = ">=3.12"
 dependencies = [
-    "narada-core==0.1.6",
+    "narada-core==0.1.7",
     # Must be a supported version in https://pyodide.org/en/stable/usage/packages-in-pyodide.html
     "packaging==24.2",
 ]
--- a/packages/narada-pyodide/src/narada/agent.py
+++ b/packages/narada-pyodide/src/narada/agent.py
@@ -168,6 +168,13 @@ class Agent(Generic[_StructuredOutput]):
                 "`reasoning` is only supported for built-in `AgentKind` values; "
                 "named Agent Studio agents own reasoning per workflow step"
             )
+        if self.model_tier != AgentModelTier.DEFAULT and not isinstance(
+            self.kind, AgentKind
+        ):
+            raise ValueError(
+                "`model_tier` is only supported for built-in `AgentKind` values; "
+                "named Agent Studio agents own model tiers per workflow step"
+            )
 
         remote_dispatch_response = await self._dispatch_request(
             prompt=prompt,
--- a/packages/narada-pyodide/src/narada/agent.py
+++ b/packages/narada-pyodide/src/narada/agent.py
@@ -51,6 +51,7 @@ from narada_core.actions.models import (
 )
 from narada_core.models import (
     AgentKind,
+    AgentModelTier,
     CriticConfig,
     File,
     McpServer,
@@ -79,9 +80,11 @@ class Agent(Generic[_StructuredOutput]):
         *,
         environment: Environment,
         kind: AgentKind | str = AgentKind.OPERATOR,
+        model_tier: AgentModelTier = AgentModelTier.DEFAULT,
     ) -> None:
         self.environment = environment
         self.kind = kind
+        self.model_tier = model_tier
 
     @overload
     async def run(
@@ -168,6 +171,7 @@ class Agent(Generic[_StructuredOutput]):
 
         remote_dispatch_response = await self._dispatch_request(
             prompt=prompt,
+            model_tier=self.model_tier,
             clear_chat=clear_chat,
             generate_gif=generate_gif,
             output_schema=output_schema,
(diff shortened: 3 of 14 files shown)


packages/narada-pyodide/src/narada/agent.py
```diff
@@ -165,9 +168,17 @@ async def run(
                 "`reasoning` is only supported for built-in `AgentKind` values; "
                 "named Agent Studio agents own reasoning per workflow step"
             )
+        if self.model_tier != AgentModelTier.DEFAULT and not isinstance(
```
xTRam1: I have a question, if someone passes the model tier as mini on the SDK, does this even change the agent steps inside the workflow? The agent step already have its own model tier thing, why would this even change anything? It makes sense that you have this error check but I want you to answer my question. What harm would it do if the user were to pass in some model tier when they were running some apa workflow command?

packages/narada-pyodide/pyproject.toml
```diff
@@ -1,14 +1,14 @@
 
 [project]
 name = "narada-pyodide"
-version = "0.1.7"
+version = "0.1.8"
```
xTRam1, line 4: Do we have to push a new change on the frontend for using the new narada pyodide package or is this done automatically? If we have to, please create a new standalone frontend PR.

packages/narada-pyodide/src/narada/agent.py
xTRam1: Correct: a top-level tier passed to a named APA workflow would not change the workflow's persisted Agent-step tiers. The harm is silent no-op behavior that suggests an override occurred. The SDK therefore rejects this combination and tells callers that named workflows own tier in their Agent steps.

packages/narada-pyodide/pyproject.toml
xTRam1: A frontend version pin is required; it is not automatic. I created the requested standalone draft PR: https://github.com/NaradaAI/frontend/pull/2066. It must remain draft until narada-pyodide==0.1.8 is published from this SDK PR.

packages/narada-core/src/narada_core/models.py
```diff
@@ -23,6 +23,13 @@ def prompt_prefix(self) -> str:
                 return "/coreAgent "
 
 
+class AgentModelTier(StrEnum):
```
xTRam1: see the comments on the backend and frontend PRs. This should be new agent type, not a model tier thing.

packages/narada-pyodide/src/narada/agent.py
```diff
@@ -79,9 +80,11 @@ def __init__(
         *,
         environment: Environment,
         kind: AgentKind | str = AgentKind.OPERATOR,
+        model_tier: AgentModelTier = AgentModelTier.DEFAULT,
```
xTRam1: for example, it will be `AgentKind.OPERATOR_MINI` for `kind`

packages/narada-core/src/narada_core/models.py
xTRam1: Fixed. The public AgentModelTier enum and all model_tier arguments were removed. Operator Mini is represented by AgentKind.OPERATOR_MINI.

packages/narada-pyodide/src/narada/agent.py
xTRam1: Implemented. AgentKind.OPERATOR_MINI emits /OperatorMini and derives the internal modelTier=mini transport field in both CPython and Pyodide.

packages/narada-pyodide/src/narada/environment.py
```diff
@@ -574,6 +576,8 @@ async def _dispatch_request(
             body["callbackHeaders"] = callback_headers
         if reasoning is not None:
             body["reasoningMode"] = reasoning.value
+        if isinstance(agent, AgentKind) and (model_tier := agent._model_tier()):
```
xTRam1: Let's not do this. Just do the regex match on the backend for /OperatorMini and set modelTier to mini there. DOn't change the request payload shapes.

packages/narada-core/src/narada_core/models.py
```diff
@@ -21,6 +22,12 @@ def prompt_prefix(self) -> str:
                 return "/Operator "
             case AgentKind.CORE_AGENT:
                 return "/coreAgent "
+            case AgentKind.OPERATOR_MINI:
+                return "/OperatorMini "
+
+    def _model_tier(self) -> Literal["mini"] | None:
```
xTRam1: Yeah delete this.

amirgholami (approved):

packages/narada-pyodide/src/narada/environment.py
xTRam1: Resolved under the final approved SDK contract. The public SDK exposes only `AgentKind.OPERATOR_MINI`; it emits `/OperatorMini` and derives the internal transport tier itself. There is no public `model_tier` argument and existing request shapes for other Agent kinds are unchanged.

packages/narada-core/src/narada_core/models.py
xTRam1: Resolved. The public `AgentModelTier` enum was deleted. The SDK now exposes only the product-facing `AgentKind.OPERATOR_MINI`, with Mini transport details kept internal.

packages/narada-pyodide/src/narada/environment.py
xTRam1: Correction after verifying the final branch: the SDK emits `/OperatorMini` only. It does not change the request payload shape or send `modelTier`; the backend command boundary selects Luna.

packages/narada-core/src/narada_core/models.py
xTRam1: Confirmed in the final branch: `AgentModelTier` is deleted. The public SDK surface is only `AgentKind.OPERATOR_MINI`, which emits `/OperatorMini`.

